PR #5, opened 2014-04-10
[imp] remove duplicated/unused folders

phproberto:
There were some duplicted folders like:

```
administrator/components/com_localise/--assets
administrator/components/com_localise/--media
```


javigomez:
Thanks @phproberto

PR #8, opened 2014-04-10
Use __construct in table

wilsonge:


mbabker:
Which Joomla version are they running this on?  If 2.5, the jimport() may
still be necessary.

wilsonge:
I checked in Skype and they said 3.3 - but I'm easy to readd it :)

phproberto:
As discussed on chat this is going to be used only on 3.x series.

PR #19, opened 2014-04-11
Cleanup formfield round 1

wilsonge:


administrator/components/com_localise/models/fields/key.php
```diff
@@ -1,17 +1,14 @@
...
-# Websites: http://joomlacode.org/gf/project/com_localise/
-# Technical Support:  Forum - http://joomlacode.org/gf/project/com_localise/forum/
--------------------------------------------------------------------------*/
-// no direct access
-defined('_JEXEC') or die;
+/**
+ * @package     Joomla.Administrator
+ * @subpackage  com_localise
+ *
+ * @copyright   Copyright (C) 2005 - 2013 Open Source Matters, Inc. All rights reserved.
```
javigomez: we are in 2014 ^_^

wilsonge: When did that happen!? :P

administrator/components/com_localise/models/fields/key.php
```diff
@@ -1,17 +1,14 @@
...
-# Websites: http://joomlacode.org/gf/project/com_localise/
-# Technical Support:  Forum - http://joomlacode.org/gf/project/com_localise/forum/
--------------------------------------------------------------------------*/
-// no direct access
-defined('_JEXEC') or die;
+/**
+ * @package     Joomla.Administrator
+ * @subpackage  com_localise
+ *
+ * @copyright   Copyright (C) 2005 - 2014 Open Source Matters, Inc. All rights reserved.
```
mbabker: You're also changing the copyright owner on this...

mbabker: Also nit-picking at this point :-P

The `Joomla.Administrator` package refers to stuff in the CMS.  For projects that aren't part of the CMS, you can use something different.

Here's how I'm doing com_patchtester - https://github.com/joomla-extensions/patchtester/blob/master/administrator/components/com_patchtester/controller.php

wilsonge: I am - we decided on Skype we're "Forking" this because this guy wasn't supposed to put his name on the copyright apparently. The majority of the other files use the OSM copyright and so I just copied and pasted that across. I think you're right we do need to change the package but it's something we should be doing across the board in one big PR.

mbabker: Got it.

On that note, at least fix the years then (and I'll stop nit-picking over small stuff for now).  It's not right that we use a blanket `2005 - <current>` statement in the CMS, it's equally wrong to assign a copyright date predating the code.

wilsonge: So just 2014 then? I mean I'm all in favour of using Phing to produce the copyright date eventually :P https://github.com/JoomJunk/Accordion/blob/development/modules/mod_accordion/tmpl/default.php#L3 + https://github.com/JoomJunk/Accordion/blob/development/build/build.properties

But OK will do. I think we just need to go through and sort out the headers through this package because it's here there and everywhere. Apparently this component predates J1.6 (which is why only parts use JForm etc) so I have no clue what the original date on the code is?

wilsonge: OK copied this out of your last PR. Better?

infograf768:
Concerning copyright, we need to include the original authors (not me, I do not care) but Christophe Demko did the biggest and Mohammad Hasani Eghtedar who continued the work after Christophe and I abandoned it.
also some parts were directly coming from/inspired by the 1.5 component by Ifan Evans.

It may not be necessary everywhere though. Let's just choose where we put that.

PR #20, opened 2014-04-14
Simplify view

wilsonge:
This will require a instance of 3.3 alpha/beta for you to test


infograf768:
Is the downloadpackage view working now?
No way I see to test

PR #21, opened 2014-04-14
My own pass on cleanups

mbabker:
- Updates the component manifest for the forked version
- Bumps the component version to `4.0.0-dev`
- First pass on cleanup of `LocaliseHelper` to include:
  - Code style with doc blocks to better identify methods purposes
  - Removes `LocaliseHelper::getActions()` in favor of `JHelperContent::getActions()`
  - Optimizes the language loading code based on the default loading en-GB handling in core
  - Drops code checking PHP versions unsupported by 3.3
  - Use less of the Filesystem package and more native PHP where appropriate

### DIFF
--- a/administrator/components/com_localise/helpers/localise.php
+++ b/administrator/components/com_localise/helpers/localise.php
@@ -1,236 +1,450 @@
 <?php
-/*------------------------------------------------------------------------
-# com_localise - Localise
-# ------------------------------------------------------------------------
-# author    Mohammad Hasani Eghtedar <m.h.eghtedar@gmail.com>
-# copyright Copyright (C) 2012 http://joomlacode.org/gf/project/com_localise/. All Rights Reserved.
-# @license - http://www.gnu.org/licenses/gpl-2.0.html GNU/GPL
-# Websites: http://joomlacode.org/gf/project/com_localise/
-# Technical Support:  Forum - http://joomlacode.org/gf/project/com_localise/forum/
--------------------------------------------------------------------------*/
-// no direct access
+/**
+ * @package     Localise
+ *
+ * @copyright   Copyright (C) 20014 Open Source Matters, Inc. All rights reserved.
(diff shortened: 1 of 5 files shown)


administrator/components/com_localise/helpers/localise.php
```diff
@@ -1,236 +1,450 @@
...
+			$vName == 'languages'
+		);
+
+		JHtmlSidebar::addEntry(
+			JText::_('COM_LOCALISE_SUBMENU_TRANSLATIONS'),
+			'index.php?option=com_localise&view=translations',
+			$vName == 'translations'
+		);
+
+		/*JHtmlSidebar::addEntry(
```
wilsonge: I know it was commented before but can we uncomment this please? As I understand it doesn't work at the moment but I guess we need to see it to work out what needs work on it :P

wilsonge:
Looks good to me at a very quick pass through. Only thing is do we want to call it v4.0.0 or as we're 'forking' it do we go to base and go as 1.0?

mbabker:
It doesn't matter as long as it's consistent.  I just went with a version
increment instead of reset.

administrator/components/com_localise/helpers/localise.php
infograf768: Please do.

infograf768: Tested after uncommenting. We will need there a lot of work. It is supposed to let create a pack for a speciific language containing specific ini files.
1: if x-debug is on, the zip contains x-debug and can't be decompressed
2. We have quite a few Undefined property and the zip produced is wrong + some deprecated calls.
3. The language choice is broken
4. The method we use to make language packs have changed since early 1.6 and we need more xmls and a package (See docs)
5. We should be able there to create specific extensions packages OR CORE packs, therefore we need to use at least the install.xml we have now in core to filter the list proposed.

mbabker: I've uncommented it.

PR #37, opened 2014-05-11
Config small corrections

infograf768:
Default permissions should be 644
.menu is an obsolete type of ini
build added to the folders to not display
Upper for string

### DIFF
--- a/administrator/components/com_localise/config.xml
+++ b/administrator/components/com_localise/config.xml
@@ -53,7 +53,7 @@
 			class="ltr"
 			required="true"
 			pattern="[0-7]{4}"
-			default="0444"
+			default="0644"
 			label="COM_LOCALISE_LABEL_SAVE_PERMISSION"
 			description="COM_LOCALISE_LABEL_SAVE_PERMISSION_DESC" />
 		<field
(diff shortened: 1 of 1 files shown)


administrator/components/com_localise/config.xml
```diff
@@ -63,12 +63,12 @@
 			description="COM_LOCALISE_LABEL_INSTALLATION_FOLDER_DESC"
 			hide_default="1"
 			default="installation"
-			exclude="administrator|component|cache|images|includes|language|libraries|logs|media|modules|plugins|templates|test|tmp|bin|cli|layouts" />
+			exclude="administrator|build|component|cache|images|includes|language|libraries|logs|media|modules|plugins|templates|test|tmp|bin|cli|layouts" />
```
javigomez, line 66: actually, the Build folder is only for the repo, it shouldn't be in the installable package.

infograf768:
Indeed, but this list exists in order to NOT display these folders when chosing an installation folder in the component configuration.
TTs may use a master instance (repo) to work on their files.
See on the same line other folders of the same type, like test, bin,cli,layouts

Issue #25, opened 2014-04-23
Move all JRequests to JInput


javigomez:
There is still some uses of JRequests, needs to be moved to JInput https://github.com/joomla-projects/com_localise/search?q=JRequest&ref=cmdform&type=Code

javigomez:
There is still a couple of places were JRequest needs to be switched to Jinput https://github.com/joomla-projects/com_localise/search?q=jrequest&type=Code

Assigning it to @gunjanpatel  for tomorrow working table

infograf768:
This has been done, closing

PR #58, opened 2014-05-30
Fixes to translation model

wilsonge:
Use JRequest plus some fixes for the doc blocks


javigomez:
Thanks! ^_^
Merging

related to #25

Issue #44, opened 2014-05-30
Fix performance issues


javigomez:
Currently the first load of com_localise starts processes all the .ini causing very poor performance. In some servers you get timeouts.

We need to study the current component and find a solution that makes it more performant. Maybe the solution could be an AJAX view that processes step by step all the things that needs to be done.

javigomez:
After installing 3 languages and trying to load the component I got this:
![screen shot 2014-05-30 at 14 02 18](https://cloud.githubusercontent.com/assets/1375475/3130134/4984e4b0-e7f2-11e3-9654-69582480c0f4.png)

javigomez:
If you reload the page about 3 times you will get the screen again :P

mbabker:
I'm thinking of the same thing with my overhauling of com_patchtester.  I was going to take the indexer from com_finder and use that as a model to create a multi-step processor.

javigomez:
Thanks @mbabker. Not sure how this indexer works. I'm really interested if you could detail the idea.

Now we have not anymore performance issues at the view thanks to @Skullbock and @fastslack fix at #76. This fix removed this calculation:
![screen shot 2014-06-03 at 13 46 01](https://cloud.githubusercontent.com/assets/1375475/3159679/86adb522-eb14-11e3-9218-0d86013bc959.png)
This calculation I think it was not that much important (do you agree @infograf768?) in exchange we solve the performance issues at administrator/index.php?option=com_localise&view=languages.

But the performance issues are still at /administrator/index.php?option=com_localise&view=translations view due to this calculations: 
![screen shot 2014-06-03 at 13 42 30](https://cloud.githubusercontent.com/assets/1375475/3159633/092f52ae-eb14-11e3-8dd0-b03e90b4c142.png)

@ianmacl mentioning you as we also spoke about the issue in case you have advise for us.

Several ideas were shared during the JandBeyond:
- the indexer of michael
- move all file calculations and strings to the database (witch I think is a lot of work and doesn't necessarily looks like the best option if we can find a way to do it with the .ini files)
- using AJAX calls after rendering the view
- @Skullbock suggested that we could try also to use pagination at /administrator/index.php?option=com_localise&view=translations to calculate only the data only for the listed items, for example for 20 items like in the example:
  ![screen shot 2014-06-03 at 13 50 52](https://cloud.githubusercontent.com/assets/1375475/3159726/32ec7f30-eb15-11e3-8e5c-bc0030b72d43.png)

[...]

PR #76, opened 2014-05-31
Removed number of files for the language view

Skullbock:
Removed number of files for the language view in order to avoid performance views
Refs #44


javigomez:
This pull enhances performance in Language view, next step is to enhance performance in translations

infograf768:
Hmm, we lost
getTotalExist and the number of files in the tip

If this is desired, then I guess the method has also to be deleted from theROOT/administrator/components/com_localise/models/translations.php

PR #81, opened 2014-05-31
Geting back Jean-Pierre changes and fixing manifest

carcam:


infograf768:
Why getting rid of the css folder and codemirror folder in the xml?
They are still present in the component

infograf768:
I will reinstate these now as they are very much needed

infograf768:
For codemirror, we can take it off if we also correct other file
PR here: https://github.com/joomla-projects/com_localise/pull/89

For css, no way: we need these

PR #82, opened 2014-05-31
Addition of a package generator function to the com_localise component

hikashop-nicolas:
The pull request adds new fields and a "download" button on the package edition interface so that you can fill the fields to be used in the language package and then then click on the "download" button in order to get a package generated on the fly and directly sent to your browser based on the settings of your package.


infograf768:
This has to be updated to the state of the branch
Please explain this patch.
How to test it? What is one supposed to get and where?
Is it for core lang files or not?

infograf768:
I managed to test this and it goes in the right direction  for core packs :)
We do have some issues though (for example the tpl ini files can't be included)
I suggest you update your branch as we can now see the default core.xml

wilsonge:
also for the header() we should probably use `JFactory::getApplication()->setHeader()`

PR #84, opened 2014-06-01
Fatal Error: cant save any ini change

infograf768:
( ! ) Fatal error: Cannot use object of type JInput as array in /Applications/MAMP/htdocs/trunkgitnew/administrator/components/com_localise/models/translation.php on line 1014


wilsonge:
my bad it should be `$input->post->getArray()`

infograf768:
is my solution OK?

wilsonge:
Yeah looks fine to me :)

PR #87, opened 2014-06-01
Search for the two packages in media folder

wilsonge:
Currently the component doesn't recognise core language files because their located in the media folder.

Apply patch and go to translations and you'll see tooltips (displayed on top of each other helpfully - UI fixes are going to be needed after this) showing the core files. 

Note that the list files is not upto date. You may wish to update them before testing. (PR https://github.com/joomla-projects/com_localise/pull/86 deals with the updates needed)


infograf768:
OK here

Issue #1, opened 2014-04-10
When creating new language, main file is not proposed for translation


javigomez:
When we create a new language, the main ini file xx-XX.ini (in admin and site) is not proposed to translation

aDaneInSpain:
I don't understand this issue. When and where is the system meant to recommend that file and should it be xx-XX.ini or should it be en-GB.ini or?

infograf768:
The PR above corrects the issue

PR #88, opened 2014-06-02
Correcting #86. Solving issue 1

infograf768:
See https://github.com/joomla-projects/com_localise/issues/1


infograf768:
Now the xx-XX. ini file is proposed to translation when it does not exist already.
Also, it can be filtered with other core files (except for installation where filter has to stay on Select Origin)

PR #89, opened 2014-06-02
Using core codemirror

infograf768:
Getting fully rid of the codemirror instance in com_localise and using core.


infograf768:
To test, make sure ypu delete the codemirror folder in media/com_localise/
Introduce an error in an ini file
For example a missing " before or after the value.
The part in error should be highlighted in red.

wilsonge:
I can't actually see where the full source form field is being used? But I can't see this breaking anything either and I support using the core version over our own version.

PR #90, opened 2014-06-02
Implementing icons for core files when necessary in the translations view. Correcting code style

infograf768:
After patch
for legend:
screen shot 2014-06-02 at 11 44 57
for body
screen shot 2014-06-02 at 11 45 57


wilsonge:
Ahh you got it working :) Just tested and it looks great.

PR #92, opened 2014-06-03
Use JRegistry instead of JObject

wilsonge:


wilsonge:
_sigh_ the one place where it does still need to be a JObject -.-

PR #93, opened 2014-06-03
Use DomDocument to create the XML package

wilsonge:
Test creating an XML package still works properly. This makes it with DomDocument which is the more 'proper' way of creating XML


administrator/components/com_localise/models/package.php
```diff
@@ -344,17 +344,28 @@ public function save($data)
...
+			$clientAttribute->value = $client;
+			$manifestElement->appendChild($clientAttribute);
+			$iconElement = $dom->createElement('icon', $data['icon']);
+			$packageSxe->appendChild($iconElement);
+			$authorElement = $dom->createElement('author', $data['author']);
+			$packageSxe->appendChild($authorElement);
+			$copyrightElement = $dom->createElement('copyright', $data['copyright']);
+			$packageSxe->appendChild($copyrightElement);
+			$licenseElement = $dom->createElement('license', $data['license']);
+			$packageSxe->appendChild($licenseElement);
```
dongilbert: In this section, I think I would do a couple things:
1) Group the element creation at the top, and the element appending at the bottom, so it's a little easier to follow.
2) Use a more descriptive var name instead of `$packageSxe`. Maybe `$packageXml`?

wilsonge: Yeah your right I guess. Sxe was because I started off using simple xml element. Then decided i didn't like it because i was then prettying the xml for the generated file at the end with dom document

wilsonge: See what you think

dongilbert:
:+1: I do like that better.

javigomez:
I like much more this way George, thanks! ^_^

Eyes hurt a bit with previous code

wilsonge:
OK merged in staging and merging

dongilbert:
Maybe this is just me, but you should never merge your own PR's.

wilsonge:
I don't see it as a problem as long as there are a number of testers. And JM went through it earlier on skype with me plus you guys did code review

dongilbert:
It's just good practice in an open source community project. If it's your own project, sure, but in that case you're probably not submitting PR's to yourself.

wilsonge:
Fair enough :) lesson learnt

dongilbert:
Thanks :D Don't think I haven't done it a time or two.

PR #94, opened 2014-06-04
Correcting some errors for core package creation

infograf768:
This saves all fields in the package xml, and lets create a core-compliant language pack.

It is not a final approach to the general packages creation where one should have the choice between a 3pd extension lang (or langs) file type pack and core.
(See http://docs.joomla.org/J2.5:Making_non-core_language_packs )
Also, for core packs, translator should have the choice to include css,  js (in media) for example

Even with this patch we still have some issues for core packs as not all ini files are included (tpl, and also some files for no apparent reason: mod_version, mod_multilangstatus, main ini file.

### DIFF
--- a/administrator/components/com_localise/models/fields/origin.php
+++ b/administrator/components/com_localise/models/fields/origin.php
@@ -66,7 +66,7 @@ class JFormFieldOrigin extends JFormField
 		$packages_options = array();
 
 		/** We took off the packages icons (due to bootstrap implementation)
-		 * This may need review
+		 * @Todo: this may need review
 		foreach ($packages as $package)
 		{
 			$packages_options[] = JHtml::_(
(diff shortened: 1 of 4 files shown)


administrator/components/com_localise/models/fields/origin.php
```diff
@@ -65,6 +65,8 @@ protected function getInput()
 		$packages         = LocaliseHelper::getPackages();
 		$packages_options = array();
 
+		/** We took off the packages icons (due to bootstrap implementation)
+		 * This may need review
```
javigomez: For code that needs future review  I would recommend to add a Todo statement:

``` php
* @Todo: this may need review
```

Because the todo statemets are indexed by the IDEs:


And we can find them easily later:

infograf768:
added

PR #98, opened 2014-06-05
Delete language link

### LINKED ISSUE #39: Remove language doesn't works
Issue at: administrator/index.php?option=com_localise&view=languages

When I click on removing a language:
screen shot 2014-05-20 at 21 27 54

I get this:

screen shot 2014-05-20 at 21 27 10

phproberto:
This fixes #39

As @mbabker commented I think using the standard checkbox + button system would alllow us to delete more than 1 language. This solves the issue and decide what to do next.

It also changes permissions for files in backend from 755 to 644


infograf768:
@phproberto 

This works great!
Do you think we could also display an alert to let know to the user that it is the whole folder for the language xx-YY in that client which will be deleted?
Too easy to make a mstake

phproberto:
@infograf768 added.

If you think we need a different string feel free to replace it

wilsonge:
This looks good. Lang. string also looks good to me

But if you delete a language then you get PHP errors in language manager because the xml file for the lang gets deleted. I dunno whether we need to uninstall the language from site/admin as well in the db so the language manager doesn't go looking for it?

infograf768:
It could indeed be an issue when a language has been installed normally (for site and admin, not for installation), as it is a pack with a manifest and, normally, can only be uninstalled globally.
For a new created language, no problem.
So, maybe we should prevent deleting the target xx-YY folder for installed packs and limit it to new languages?

Also, I think we should limit the deletion to one language only at a time, therefore the way it is now done is OK for me

infograf768:
This is what I suggest to prevent deleting installed packs:
in /administrator/components/com_localise/models/language.php
add line 435 something like

```
        $db = JFactory::getDbo();
        $query = $db->getQuery(true)
            ->select('name')
            ->from($db->quoteName('#__extensions'))
            ->where($db->quoteName('type') . ' = ' . $db->quote('package'))
            ->where($db->quoteName('state') . ' = 0')
            ->where($db->quoteName('element') . ' = "pkg_' . $tag . '"');
        $db->setQuery($query);
        $installedPack = $db->loadResult('name');

        if ($installedPack != null)
        {
            $this->setError(JText::sprintf('COM_LOCALISE_CANNOT_REMOVE_INSTALLED_LANGUAGE', $tag));

            return false;
        }
```

infograf768:
The PR has to be updated to the branch

Issue #99, opened 2014-06-06
Document testing


javigomez:
It would be awesome if you @zanardigit or @puneet0191 could create here: https://github.com/joomla-projects/com_localise/tree/develop/tests/system a readme.md file that explains step by step how to run the tests. Something like we have at http://docs.joomla.org/Running_Automated_Tests_for_the_Joomla_CMS but smaller, just with a few steps to set up.

zanardigit:
Done, PR sent.

javigomez:
Thanks, adding here the part mentioned by @puneet0191  in the pull request: https://github.com/joomla-projects/com_localise/pull/100#issuecomment-45406055

To close this issue we need some one to document how to configure the configuration file. By the way I think we forgot to add the configuration.php.dist example file like we have in joomla cms (https://github.com/redCOMPONENT-COM/redSHOP/blob/develop/tests/system/servers/configdef.php.dist) I don't see we have any in the repository

cc: @zanardigit

puneet0191:
yes, we forgot to add that file..and isn't the configuration file self explainatory? or do we need a doucment to explain it?

javigomez:
yes, the configuration file is self-explanatory, just mention in document were the people can find the configuration file

puneet0191:
okay, will update the document with details of configuration file

PR #100, opened 2014-06-07
readme.md

zanardigit:
Minimal version of a readme.md to explain how to use the webdriver test suite.


puneet0191:
@zanardigit Thanks :), I think we need to add the part where users are supposed to make changes in the configuration file and they are also required to update git submodule before they begin running system tests

javigomez:
Related to issue #99

javigomez:
Merging and adding Puneet mentioned task at issue #99

PR #101, opened 2014-06-07
[CODE STYLE] Fixing coding style in main files

gunjanpatel:


administrator/components/com_localise/localise.php
```diff
@@ -1,6 +1,7 @@
 <?php
 /**
- * @package     Com_Localise
+ * @package     Joomla.Administrator
```
wilsonge: This is still the com_localise package - same for the script below

PR #102, opened 2014-06-08
Coping with some Core inis not included in reference

infograf768:
Some core ini files are not included in reference.
In the translations view, they display as not in reference for an already installed language.
When it is a new language, they do not even display
here is an example with the com_wrapper ini and sys.ini in administrator
Before:
screen shot 2014-06-08 at 11 08 55

After:
screen shot 2014-06-08 at 10 52 38

I have traced this issue to the existence or not of the component folder in the $client, the way the scans are done, i.e. if I add an empty com_wrapper folder in admin, then these files wou;d be included as reference automatically.

The PR forces them as reference.

### DIFF
--- a/administrator/components/com_localise/models/translations.php
+++ b/administrator/components/com_localise/models/translations.php
@@ -360,6 +360,37 @@ class LocaliseModelTranslations extends JModelList
 				{
 					foreach ($files as $file)
 					{
+						// Coping with Core files not considered as reference
+						if ($file == $reftag . '.com_messages.ini' || $file == $reftag . '.com_mailto.sys.ini'
+							|| $file == $reftag . '.com_wrapper.ini'|| $file == $reftag . '.com_wrapper.sys.ini')
+						{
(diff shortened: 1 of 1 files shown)


javigomez:
I'm not sure to understand what "included for reference means", could you explain? Thanks

infograf768:
Some ini files may be present for example in the fr-FR language folder, but they do not exist in the language which is used as reference for the translation (most of the time en-GB).
Usually these files would not be included in a Core pack.

Special case these days: geshi.
We took off the plugin from our distributions, and also the language files.
Lang communities distributions keep including geshi ini files for B/C
They will look as "Not in reference" as they are not present in the en-GB folder.
That case is totally normal.

In the cases treated in this PR, the core files concerned should not be considered as "Not in reference" as they do exist (and as core) in the en-GB folders.
The files concerned are:
com_messages.ini 
com_mailto.sys.ini 
com_wrapper.ini 
com_wrapper.sys.ini

PR #105, opened 2014-06-15
*Allowing the creation of File packages + other improvements

infograf768:
The component needs to be able to create Core lang packs or extensions file lang packages.
See model here: http://docs.joomla.org/J2.5:Making_non-core_language_packs
We now can make both:
screen shot 2014-06-15 at 09 19 31

For the moment it only does for one language at a time.
To test, install a default 3.3.1 and install a core lang pack
Open the Translations view and filter by Origin: Core to see the state of the files and populate the database
Go to Packages and choose new Core pack
choose all files that should go in (for example do not add com_localise files)
for the version enter 3.3.1.1, the language concerned and save, Without quitting this edit page click on the download button.

Install a third party component (for example patchtester)
Translate in the Translations view for the same language you installed previously
Go to packages
Click New File Package and choose the 2 patch tester files

Click on download

The PR also corrects/improve other aspects, like filtering by Core files in translations view 
(@Todo: 
Add xx-XX.finder_cli.ini which does not follow "normal" file names.
Allow creation/editing of the xx-XX.localise.php, by adding it to the view and letting copy from reference in a codemirror editable instance.
Allow creation/editing of the xx-XX.css, by adding it to the view in a codemirror editable instance.
Allow, when making a core pack, adding some stuff in the media folder (calendar for Farsi is a good exemple)

All code style is not perfect and I have left a lot of older code in.
This should be considered as a step in the refactoring process.


administrator/components/com_localise/controllers/packagefile.php
```diff
@@ -0,0 +1,185 @@
...
+	/**
+	 * Method to check if you can add a new record.
+	 *
+	 * Extended classes can override this if necessary.
+	 *
+	 * @param   array  $data  An array of input data.
+	 *
+	 * @return  boolean
+	 */
+	protected function _allowAdd($data = array())
```
javigomez, line 89: According to our coding standards we don't use underscore for defining a private class property or method

``` php
// Please in future pulls fix this:
protected function _allowAdd

// for:
protected function allowAdd
```

administrator/components/com_localise/models/packagefile.php
```diff
@@ -0,0 +1,743 @@
...
+	 * Method to auto-populate the model state.
+	 *
+	 * Note. Calling getState in this method will result in recursion.
+	 *
+	 * @param   string  $ordering   An optional ordering field.
+	 * @param   string  $direction  An optional direction (asc|desc).
+	 *
+	 * @return  void
+	 *
+	 * @since   1.6
```
javigomez, line 36: I think this since can be removed

javigomez:
Tested, excelent Job @infograf768. I like a lot the idea of having diferentiated buttons for the different types of packages.
I see as you said that only works with one file. As soon as we make it valid for more than one language files I suggest that we rename this button from:

screen shot 2014-06-16 at 11 26 28

to:

screen shot 2014-06-16 at 11 29 09

@infograf768 I think we can have a **array with all the files that are from core**, then if you do a "extension package" we can filter the file list field to only show this files that are not in the  **array with all the files that are from core** so we don't show this looooooong list:

screen shot 2014-06-16 at 11 22 03

Also maybe we can use this modern bootstrap multiselect with search box and checkboxes, something like:

screen shot 2014-06-16 at 11 36 39

@phproberto do you know if we can easily do this with chosen or select2?
I mean a multiselect form with search box and checkboxes. Actually I think this one will work: http://harvesthq.github.io/chosen/#optgroup-support

infograf768:
With this PR, Javier, you will not get anymore these ini of type fil_localise_etc as this was commented out

Concerning coding standards, you may have remarked that I basically copied some existing code to create the new packagefile.suffix, therefore we do need to correct that kind of stuff also in package.suffix, as well as many code standards stuff... 

Concerning " I see as you said that only works with one file.", I rather meant "only for one language at a time" as we already can include many files in the filepackage now (com, mod, lib, plg) for site and/or admin. No issue calling this button New Extension Package.

In fact, the first iteration of com_localise coded by Christophe Demko did allow to load specific xml per extension and save/share them.

Would be good indeed to be able to take off all defined core files from the file package choice.

javigomez:
you said:

<blockquote>
Concerning " I see as you said that only works with one file.", I rather meant "only for one language at a time" as we already can include many files in the filepackage now (com, mod, lib, plg) for site and/or admin. No issue calling this button New Extension Package.
</blockquote>

when I tested adding more than one file to the "file package" I get an error. The package is not saved and returns back to the empty form without populating any data. I thought this was work in progress.

infograf768:
You may have not done what should.
All the files concerned have to be saved first as Translations in the target language.
here is an example with some com_ and a mod_:
screen shot 2014-06-17 at 08 17 34

here is the xml

```
<?xml version="1.0" encoding="UTF-8"?>
<extension type="file" version="1.0" method="upgrade">
    <name>tester_plus_stoolfr-FR</name>
    <version>1.0.0</version>
    <creationDate>17/06/2014</creationDate>
    <author>me</author>
    <authorEmail>me@mail.com</authorEmail>
    <authorUrl>http://asite.com</authorUrl>
    <copyright>my copyright</copyright>
    <license>http://www.gnu.org/licenses/gpl-2.0.html GNU/GPL</license>
    <description><![CDATA[a test file pack]]></description>
    <fileset>
        <files folder="site/fr-FR" target="language/fr-FR">
            <filename>fr-FR.mod_stoolpigeon.ini</filename>
    </files>
        <files folder="admin/fr-FR" target="administrator/language/fr-FR">
            <filename>fr-FR.com_patchtester.ini</filename>
            <filename>fr-FR.com_patchtester.sys.ini</filename>
        </files>
    </fileset>
</extension>
```

We have to improve this by displaying a message if the file desired does not exist in the target language.

PR #107, opened 2014-06-16
removing useless method

javigomez:
all this methods can be removed. They are useless due inheritance


infograf768:
Merged. Thanks.
Removing also the package.php one directly

PR #108, opened 2014-06-16
Delete localise assets using JTable

fastslack:
Each time the # __localise table is generated with the translations data files, also adds a corresponding row in # __assets table. Currently in the purge function, we are deleting only the rows in the # __localise table, but not into # __assets. This results in a significant loss of performance not only in this component, but throughout the installation of Joomla!.


infograf768:
Corrected some code styles and merged. 
Thanks!

NB: To test com_localise further, install a new Joomla and com_localise instance as it takes ages to delete all com_localise ancient assets

javigomez:
gracias @fastslack ^_^

Issue #64, opened 2014-05-31
In translations view Icon button always appears


carcam:
In translation views Icon button always appears even though icon does not appear because it's not needed.
![selection_050](https://cloud.githubusercontent.com/assets/1695138/3139181/518fe638-e8cd-11e3-8e96-c2ec40f38bee.png)

Issue #70, opened 2014-05-31
Fix columns in translations view


carcam:
In translations view (administrator/index.php?option=com_localise&view=translations) replace current columns and icons with this structure:

Tag | Translation name | Translated | Storage / Path | Strings / Phrases | Author | Type | info

Tag: Language tag
Translated: Percentage of wokr done
Storage/Path: Path of the language file
Strings /Phrases: Number of strings language files contains
Author : Whose the author of the lnaguage file and it would be nice to include an icon if it's Joomla core
Type: Component, module, plugin, library...
Info: An icon with tooltip for extra information

infograf768:
The info is presented first. I think we now have quite a good display.
Adding the path as a sub to the view. See PR

PR #111, opened 2014-06-17
In translations view Icon empty button always appears

infograf768:
see: https://github.com/joomla-projects/com_localise/issues/64

we will now get the icon only when necessary, for example:
screen shot 2014-06-17 at 18 11 24

Also some code style in this PR


infograf768:
Adding the path to the file.
See: https://github.com/joomla-projects/com_localise/issues/70

screen shot 2014-06-17 at 18 31 17

phproberto:
Looks good here! :+1:

PR #117, opened 2014-06-19
Throwing errors and preventing creating core pack if files missing.

infograf768:
Also creates a basic xx-XX.localise.php if absent.
This last one can be changed if/when we enable the creation of a custom localise.php in one of the views.


infograf768:
Examples:
If the basic xx-XX.xml does not exist
screen shot 2014-06-19 at 11 27 34

if some files are missing
screen shot 2014-06-19 at 11 32 41

javigomez:
nice! Thanks.

PR #124, opened 2014-06-20
Set headers and exit with JApplication

wilsonge:


javigomez:
nice! thanks ^_^

infograf768:
This has broken downloading file package zip

wilsonge:
OK I don't have time this weekend to work on a fix. Not free until monday. if you want to revert I'll have another shot at it :/

javigomez:
Ops, it worked to me when I tested. What error are you having JM?

wilsonge:
He commented on the commit itself. Sorry I got a friend round all weekend so no coding time available

infograf768:
I better revert this to be able to use the functionnality until new code.
@javigomez 
It is the packagefile download which is broken, not the core package one.

mbabker:
This doesn't work as is because the headers are never sent.  Calling `$app->close();` is a proxy to exit basically.

In all actuality, you should never need to close the app before the full cycle completes, but in our environment, there's too much bad code floating around and in non-HTML response types, it's easier to just close the app than let a plugin screw things up.

wilsonge:
So you think that if we just immediately call `$app->redirect` to the zip file that will work? I just replaced the exit with `$app->close()` as `exit` was used previously.

mbabker:
Redirecting is no better as it'll just trigger another page load.  If the specific request that's calling this code is designed to serve a download, make sure it isn't serving up HTML.  So whatever is calling it should probably include `&format=raw` in the URL and there not be any exit/close calls in the process, that'd be the closest answer to right.  You're just having to work around design issues that make the CMS inefficient in areas like this; the application shouldn't be coupled to JDocument for all requests.

wilsonge:
So the function initially creates the zip then it should download it. it's never html code.

mbabker:
Looking at Akeeba Release System, it's no better in that it too is just calling `header()` directly then closing the application.  But, there is some code that could be borrowed from it regarding download chunks and whatnot to help with file sizes and slower connections.

Calling `$app->setHeader()` just buffers the headers until the application gets to the [response step](https://github.com/joomla/joomla-cms/blob/master/libraries/joomla/application/web.php#L409-L450) whereas `header()` starts sending the headers and sets them immediately.  So calling exit basically keeps the application from doing its job.  Within our system, there's too much chance for plugins to screw things up on non-HTML pages, so you do see a lot of developers exiting the application early to get around that.  Bad catch-22, but I'm pretty willing to bet that there wasn't a lot of thought about the CMS serving non-HTML content, let alone file downloads, back in the old days.

phproberto:
My 2 cents: 

A model should never deal with application close or exit. It should provide a zip file. It's the controller who has to deal with headers and close / exit.

javigomez:
Ok, reverting, while we find the solution.

javigomez:
Oh! JM was faster https://github.com/joomla-projects/com_localise/commit/5acb9caa1fcf1c69d1cf2708100bc3ca5b5849ad ^_^

wilsonge:
The revererting still ends up with us exiting :P there should be another issue for this. My friend leaves this afternoon so will have a look over it then and see what I can do

PR #125, opened 2014-06-21
Limiting language choice when creating packages

infograf768:
Limiting language choice to existing core site and administrator xx-XX.xml when creating packages. Cleaning up 'installation' related code for package.php.

Before:
screen shot 2014-06-21 at 07 10 39

After
screen shot 2014-06-21 at 07 08 25

### DIFF
--- a/component/admin/models/fields/corelanguage.php
+++ b/component/admin/models/fields/corelanguage.php
@@ -1,0 +1,85 @@
+<?php
+/**
+ * @package     Com_Localise
+ * @subpackage  models
+ *
+ * @copyright   Copyright (C) 2005 - 2014 Open Source Matters, Inc. All rights reserved.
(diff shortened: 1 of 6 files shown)


component/admin/models/fields/corelanguage.php
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Renders a list of all languages
+ * Use instead of the joomla library languages element, which only lists languages for one client
+ *
+ * @package     Extensions.Components
+ * @subpackage  Localise
+ *
+ * @since       1.0
+ */
+class JFormFieldCoreLanguage extends JFormField
```
javigomez, line 23: I think this can be a JFormFieldList instead and do a getOptions instead of doing a getInput but looks okay, so merging and creating an issue for future

javigomez, line 23: see #127

Issue #130, opened 2014-06-22
fix code style issues in all files


javigomez:

infograf768:
I guess we can close this now as Jenkins and Code Sniffer are happy

PR #131, opened 2014-06-22
Fix only code style issues in language and package models

javigomez:
related to issue #130

### DIFF
--- a/component/admin/models/package.php
+++ b/component/admin/models/package.php
@@ -327,7 +327,7 @@ class LocaliseModelPackage extends JModelForm
 		$package  = $this->getItem();
 		$path     = JPATH_COMPONENT_ADMINISTRATOR . "/packages/$name.xml";
 		$manifest = $name;
-		//$client   = $package->client ? $package->client : 'site';
+		// $client   = $package->client ? $package->client : 'site';
 
 		if ($package->standalone)
 		{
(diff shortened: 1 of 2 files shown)


infograf768:
Thanks

component/admin/models/package.php
```diff
@@ -344,7 +345,6 @@ public function save($data)
 			$manifestElement = $dom->createElement('manifest', $manifest);
 			$versionElement = $dom->createElement('version', $data['version']);
 			$authorElement = $dom->createElement('author', $data['author']);
-			$copyrightElement = $dom->createElement('copyright', $data['copyright']);
```
wilsonge, line 347: Should this have been removed?

javigomez, line 347: yes, is repeated just a few lines after. It was ovewritten.

javigomez, line 347: specifically here https://github.com/joomla-projects/com_localise/pull/131/files#diff-aabe745fa672e9b0deb72a3c70e71525R352

wilsonge, line 347: Oh yeah :) my bad!

javigomez, line 347: no no, your good. Is super great that we review others pulls ^_^

wilsonge, line 347: https://github.com/joomla-projects/com_localise/pull/137 we should do this then as well for same reason

PR #136, opened 2014-06-22
remove not used helper file

javigomez:
I was going to rearrange this helper file into the /helpers/ folder but after digging a bit I have reached the conclusion that we are not using it anywhere. I have searched calls to the class and to the methods and I can't find it anywhere.

In this pull I'm removing it, and after testing the removal I have tested al general functionality without issues


infograf768:
As I guess the code is now in the translation model, I merged. Thanks

PR #139, opened 2014-06-22
Refactor to use a PHP script for PHPCS on Travis

mbabker:
PHPCS does not cause a Travis build to fail when there are codestyle errors detected, making a build appear to pass all tests.  To get around this, a custom script must be used to run PHPCS and flag the build script as failed.

This will use Composer to install PHPCS in place of the PEAR dependencies on PHPCS and Phing in the Travis environment.  The documentation is also updated to reflect that PHPCS can be run using Ant, Phing, or this PHP script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,10 @@ build/logs
 /tests/system/servers/configdef.php
 /tests/system/webdriver/tests/logs/
 
+# Composer
+composer.lock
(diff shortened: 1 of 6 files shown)


javigomez:
Wow! this is super interesting. Thanks @mbabker ^:^

PR #140, opened 2014-06-22
remove unneeded method due to inheritance

javigomez:
This is part of #130

Removes an unneeded method due to inheritance

### DIFF
--- a/component/admin/controllers/language.php
+++ b/component/admin/controllers/language.php
@@ -20,20 +20,6 @@ defined('_JEXEC') or die;
 class LocaliseControllerLanguage extends JControllerForm
 {
 	/**
-	 * Method to get a model object, loading it if required.
-	 *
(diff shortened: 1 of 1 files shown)


infograf768:
javier, I had reverted this from your PR because we could not delete a language.xml anymore

Warning

JFolder: :delete: Path is not a folder. Path: /language/
Error

Error removing languages: /language/

infograf768:
Same issue with translation...
https://github.com/joomla-projects/com_localise/pull/144

javigomez:
mmm, interesting. I will dig on it. Thanks for the warning.

phproberto:
Removing getModel functions disconnects models and controllers so any standard action will be broken. 

My advice: if model exists never remove the method.

infograf768:
Closing for now.

PR #146, opened 2014-06-22
fix code style in package model

javigomez:
related to #130


component/admin/models/package.php
```diff
@@ -770,8 +778,6 @@ public function download($data)
 
 			// Generate administrator package
 			$admin_package_files = array();
-			$admin_zip_path = JPATH_ROOT . '/tmp/' . uniqid('com_localise_') . '.zip';
```
javigomez, line 773: note for reviewer: this variable is ovewrided some lines after with the same value

PR #155, opened 2014-06-27
Core packs need specific version format

infograf768:
We do need to get for core packs a version of type 3.3.1.1 where the last figure is the pack specific version for Updates. See registered language packs
Thanks Brian for the idea to separate the J version from the pack version to clarify the UI.
(NB: Alpha ordered that part of the ini file.)

One will get:
screen shot 2014-06-27 at 08 59 58


component/admin/language/en-GB/en-GB.com_localise.ini
```diff
@@ -348,52 +348,56 @@ COM_LOCALISE_ERROR_PACKAGE_WRITABLE="Could not make the package file %s writable
...
+COM_LOCALISE_LABEL_PACKAGE_JOOMLA_VERSION="Joomla! version"
+COM_LOCALISE_LABEL_PACKAGE_JOOMLA_VERSION_DESC="Enter the Joomla! version. Example: 3.3.1, 3.3.2, 3.4.1, etc."
+COM_LOCALISE_LABEL_PACKAGE_LANGUAGE="Language"
+COM_LOCALISE_LABEL_PACKAGE_LANGUAGE_DESC="Language"
 COM_LOCALISE_LABEL_PACKAGE_LICENSE="Licence"
 COM_LOCALISE_LABEL_PACKAGE_LICENSE_DESC="Licence description"
 COM_LOCALISE_LABEL_PACKAGE_NAME="Name"
 COM_LOCALISE_LABEL_PACKAGE_NAME_DESC="Name description"
+COM_LOCALISE_LABEL_PACKAGE_PACK_VERSION="Pack version"
+COM_LOCALISE_LABEL_PACKAGE_PACK_VERSION_DESC="Enter the pack version. Exemple: 1, 2, ... 12"
```
brianteeman: This should be example not exEmple

brianteeman:
What do you think about changing the words "pack version" to "Language Version"? I think  thats more descriptive of what it actually is

wilsonge:
If the other field is just joomla version then can it just be taken from JVersion? Or are we still going to be using this component to generate 2.5 packs etc?

infograf768:
What happens is that a package can be made when JVersion is not the one targetted at.
For example, when one prepares 3.3.2 but on a 3.3.1 site.

wilsonge:
OK gotcha. I guess we could make JVersion the default value of the field. But it seems a bit excessive making a custom form field just for that :/

PR #156, opened 2014-06-27
Packages can't be deleted

infograf768:
( ! ) Fatal error: Call to a member function delete() on a non-object in /Applications/MAMP/htdocs/trunkgitnew/administrator/components/com_localise/controllers/packages.php on line 83


infograf768:
Closing this after reinstating the getModel in the controller

Issue #135, opened 2014-06-22
*Errors when saving new package


infograf768:
We have 3 errors when saving a new package
1. PHP Notice:  Undefined index: id in ROOT/libraries/legacy/controller/form.php on line 753
2. After save, we are redirected to a new empty form/page
3. Therafter, when clicking on New Core Pacakage or New Extension Package, the page loaded is not empty but filled with the content of the last package saved.

To avoid the last error, I have to quit my browser (Firefox) and restart it.

infograf768:
https://github.com/joomla-projects/com_localise/issues/158
 should take care of issues 2 and 3

infograf768:
Remains 1. PHP Notice: Undefined index: id in ROOT/libraries/legacy/controller/form.php on line 753

infograf768:
Last Notice solved by https://github.com/joomla-projects/com_localise/pull/210

PR #157, opened 2014-06-29
Errors when saving new package (Issue #135)

infograf768:
Errors when saving new package 
https://github.com/joomla-projects/com_localise/issues/135

This should take care of the errors, except for the Notice


infograf768:
Please use
https://github.com/joomla-projects/com_localise/pull/158

PR #158, opened 2014-06-29
Errors when saving new package (Issue #135)

infograf768:
replaces PR https://github.com/joomla-projects/com_localise/pull/157

### DIFF
--- a/component/admin/controllers/package.php
+++ b/component/admin/controllers/package.php
@@ -33,12 +33,12 @@ class LocaliseControllerPackage extends JControllerForm
 
 		// Get the id
 		$cid = $input->get('cid', array(), 'array');
-		$cid = count($cid) ? $cid[0] : '';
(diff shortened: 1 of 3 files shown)


Achal-Aggarwal:
@test
After applying the patch I am able to create new packages (core | extension) after saving a new package.
It also solves issue 3 but not 2 mentioned in #135 
Form is getting empty after save, it is like "save and new" instead of just "save".

Another issue I noticed was
4. when the newly created packages is deleted then creating a new package throws error.
screenshot from 2014 07 01 23 18 14

This patch also solves this issue, so test this one also.

infograf768:
The remaining issue I have here is only for New Extension pack, where the fields are emptied after Save, indeed as if it was a "Save and New"

No issue for New Core pack

Achal-Aggarwal:
@test
fields retain their value on save for "New Extension package".

Issue #126, opened 2014-06-21
Add chosen in Extension Package view


javigomez:
We need to add CHOSEN in the Extension Packager

![screen shot 2014-06-21 at 09 31 36](https://cloud.githubusercontent.com/assets/1375475/3348142/ed651dfa-f915-11e3-955e-e6006fd5c9e9.png)

This is the field that needs to be replaced:

![screen shot 2014-06-21 at 09 31 16](https://cloud.githubusercontent.com/assets/1375475/3348140/e26c5242-f915-11e3-8c4c-3687345343da.png)

Is a field called /fields/translations.php. 
We cant' remove it because "core package is using it". I think we should create a new one because the functionality will be different, somethign like: extensionstranslations.php

This new field needs to have:
- [ ] search box to easily find the extension files, like chosen allows:
  ![screen shot 2014-06-21 at 09 35 21](https://cloud.githubusercontent.com/assets/1375475/3348146/77dbf922-f916-11e3-8b8f-95b61e507059.png)
  - [ ] The list should show only extensions that are not core extensions. So we need to remove from the list these extensions witch are not listed in /media/com_localise/packages/core.xml

infograf768:
I also suggest to be able to import existing xmls for a specific extension/group of extensions where all the necessary files and only these will be listed.

infograf768:
Followed your suggestion by creating an extensiontranslations.php field

infograf768:
I tested adding
JHtml::_('formbehavior.chosen', 'select');

in the packagefile/edit.php view
this is what I get:
before clicking on the field (here is an already made pack)
![screen shot 2014-10-06 at 12 25 17](https://cloud.githubusercontent.com/assets/869724/4524368/4fd30610-4d43-11e4-9e5f-abece42e101b.png)

after clicking in the field
![screen shot 2014-10-06 at 12 25 42](https://cloud.githubusercontent.com/assets/869724/4524370/5ba931da-4d43-11e4-9ee6-3b6bec6d2eb1.png)

What you think?

wilsonge:
Ummm I don't think it looks too bad like that although I worry how it will look with like 20/30 translation files selected for a larger package (take akeeba or something when you have a load of plugins to accompany a component).

Also we should always attach chosen to a class rather than all select elements.

PR #160, opened 2014-06-30
Limiting extension package choice to non-core files

infograf768:
Partly solving https://github.com/joomla-projects/com_localise/issues/126 as no chosen.
When creating a new Extension package, the choice will not include core files.
We will get for example:
screen shot 2014-06-30 at 17 08 56


Achal-Aggarwal:
@test
getting grouped list without core files, as expected.
If group is empty can we add an empty option `<option value="">No translation (or something like this)</option>`

infograf768:
@Achal-Aggarwal 
Good idea.

infograf768:
Adding now the new string directly

PR #163, opened 2014-07-02
Implementing Export of package xml

infograf768:
This should complete the Import xml. It lets share localise package xml between users to allow Import and use on another box.
screen shot 2014-07-02 at 18 29 45


infograf768:
FYI, both the core.xml and the localise.xml will NOT be exported as they are considered as reference languages.

We may not need the localise.xml as is (it should be comparable to the packages we create in com_localise/packages  and mayne moved there) and I am not sure thatcore.xml should show in the packages view.

I think instead, that we should take off the readonly for the packages title/name, and implement a Save and Copy for any pack in order to use already made packs for one specific lang where the files desired are already highlighted and modify the copy to fit the target language.
In the case of core.xml, we would add a new correctly formatted new file in com_localise/packages/

Achal-Aggarwal:
@test
I have successfully exported/downloaded the xml of newly created core/extension pack.

Issue #165, opened 2014-07-04
TODO: Implement Save and copy/duplicate packages


infograf768:
This will let use a master package xml as a model for new package

PR #167, opened 2014-07-05
Implementing package duplication

infograf768:
See https://github.com/joomla-projects/com_localise/issues/165

This will let use master packages to create a specific language package where the files to include are already highlighted.
screen shot 2014-07-05 at 11 34 52


infograf768:
Screencapture with time stamp:
screen shot 2014-07-06 at 07 37 04

PR #168, opened 2014-07-06
Implementing duplicating packages

infograf768:
Replaces https://github.com/joomla-projects/com_localise/pull/167

See #165

This will let use master packages to create a specific language package where the files to include are already highlighted.
screen shot 2014-07-06 at 07 37 04


javigomez:
Tested, works:

screen shot 2014-07-06 at 17 34 09

Thanks ^_^

PR #169, opened 2014-07-06
Adding PHP Mess Detector to travis

javigomez:
Adds PHP Mess Detector, that will give us automated interesting advises to improve the code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,4 @@ before_script:
 script:
   - find ./component -name '*.php' -exec php -l {} \;
   - php .travis/phpcs.php
+  - ./vendor/bin/phpmd component/ text codesize,unusedcode,naming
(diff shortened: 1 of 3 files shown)


wilsonge:
So I think there are a few issues with this. Mainly that it makes travis fail because we have unused variables (which is always gonna happen with the mvc structure of joomla (new or legacy)). Plus `$id` is not at least 3 chars long which is just barmy

http://docs.travis-ci.com/user/build-configuration/ This talks about allowing things to fail and it would be nice if we could have this but allow it to fail. So we have the pointers but don't have to implement all of them.

javigomez:
I have removed the Naming checker http://phpmd.org/rules/naming.html

Actually we can create our own ruleset: http://phpmd.org/documentation/creating-a-ruleset.html

javigomez:
maybe we can convert the call into a PHP script that returns the output and exits with 0

Something like Michael did with PHPcs

javigomez:
Now works:

screen shot 2014-07-06 at 15 29 57

...

screen shot 2014-07-06 at 15 30 08

^_^

javigomez:
Thanks @wilsonge  ^_^

PR #175, opened 2014-07-07
[imp] build improvements

phproberto:
This moves files:

```
localise.xml
install.php
```

to the root folder of the project making now possible:
- Install from folder using project root
- Install from folder user project/build/packaging after building package
- Install from generated package file after building package
- Discover install on test site after running `phing -f copier.xml`
- Download & install generated package from github

So basically all the installation types should work.


javigomez:
Adding in cc to @puneet0191 and  @kshitijSharma2014 because System tests will be affected.

puneet0191:
we will have to update the Installation script now, but waiting for FB-webdriver to come in action :)

Issue #174, opened 2014-07-07
TODO: Master/reference packages, core.xml


infograf768:
In the original com_localise (before 1.6), a difference was made between packs.
Master/reference packs were placed in media/com_localise/packages
User made packs in administrator/components/com_localise/packages/

A specific naming was used "fil_etc.
Also a variable: $standalone

Along the years and modifications, this got lost. 
the media/com_localise/packages/localise.xml does not make sense at all for example and should be modified.

We should reinstate a differenciation so as to complete the code introduced to import/export/duplicate.

Also: our core reference file, core.xml, should only be a reference, as its formatting can't be used to produce a final core package. 
Therefore, I think it should not show in the packages view and should never be exported, deleted or even duplicated. 
Instead we could have a basic core package as produced by creating one through "New Core package" which should be used to be cloned/duplicated by TTs to adapt to their language.

Will make a PR to try to implement that aspect, (and modify the existing localise.xml)

In both case, we have to devise a way to update these depending on the core install.xml files for en-GB in the main en-GB folders.

==> NB: we still need to work on 
https://github.com/joomla-projects/com_localise/issues/15
https://github.com/joomla-projects/com_localise/issues/7
https://github.com/joomla-projects/com_localise/issues/6

PR #178, opened 2014-07-09
*Creating master_ core and master_localise packages in the component packages, separating master from user 

infograf768:
See https://github.com/joomla-projects/com_localise/issues/174
This PR prevents the core.xml file to show in the packages view, keeping it as a reference only.

It adds 2 master_ packages in the component packages folder with the ini files already highlighted and prevents generating and downloading installable packs from them (if someone tries, an alert will explain that the master pack should first be duplicated and customised for the language)
The localise.xml that was in the media folder is deleted.

Basically, any package created which contains "master_" in the name of the file will behave that way and can be shared, as well as the non-master ones.

Also is implemented an alert when importing an xml package to tell the user that this import may overwrite an existing file. George tried to throw that alert only when a file with same name existed but we have errors (see https://github.com/infograf768/com_localise/pull/2 )

Here is an example of what we get in the manager:
It should be improved in a next step to include a new column to show master packages.
screen shot 2014-07-09 at 17 54 43


component/admin/packages/master_core.xml
```diff
@@ -0,0 +1,305 @@
...
+    <filename>plg_user_joomla.ini</filename>
+    <filename>plg_user_joomla.sys.ini</filename>
+    <filename>plg_user_profile.ini</filename>
+    <filename>plg_user_profile.sys.ini</filename>
+    <filename>tpl_hathor.ini</filename>
+    <filename>tpl_hathor.sys.ini</filename>
+    <filename>tpl_isis.ini</filename>
+    <filename>tpl_isis.sys.ini</filename>
+  </administrator>
+</package>
```
javigomez, line 305: If you move this to /admin/packages can we remove https://github.com/joomla-projects/com_localise/blob/develop/media/com_localise/packages/core.xml ?

infograf768, line 305: No, that is the whole point: we need a core.xml as a reference without the .ini suffix
When I tested with the ini, I lost the core icon in translations view.

Issue #180, opened 2014-07-11
Changing name in package generates new


javigomez:
See the video:

http://www.youtube.com/watch?v=scXvbsNjidY

Instead of renaming a package it generates a new one.

infograf768:
I confirm the issue

infograf768:
I think it is due to
https://github.com/joomla-projects/com_localise/pull/164
Before that, the field was readonly

The issue here is that we need to be able to change the name, specially when duplicating

infograf768:
Tanks for testing the PR

PR #183, opened 2014-07-15
 Changing name in package generates new package:  PR for #180 

infograf768:
This should solve https://github.com/joomla-projects/com_localise/issues/180 by deleting the older file when changing the name of the xml


component/admin/models/package.php
```diff
@@ -313,6 +313,22 @@ public function getItem()
 	 */
 	public function save($data)
 	{
+		// When editing a package, find the original path
+		$app = JFactory::getApplication('administrator');
+		$this_id = $app->getUserState('com_localise.edit.package.id');
```
phproberto: Can we call this `$id` following joomal standard?

component/admin/models/packagefile.php
```diff
@@ -311,6 +311,22 @@ public function getItem()
 	 */
 	public function save($data)
 	{
+		// When editing a package, find the original path
+		$app = JFactory::getApplication('administrator');
+		$this_id = $app->getUserState('com_localise.edit.package.id');
+
+		if (!empty($this_id))
```
phproberto: Same here please `$id`

infograf768:
Updated

javigomez:
Works!

screen shot 2014-07-17 at 11 57 56

Thanks @infograf768 

This pull closes  #180

PR #182, opened 2014-07-14
[fix] Build system. Fixes #181

phproberto:
This fixes build calls from outside the build folder. Now we have:
### From extension root

``` bash
phing -f build/build.xml
```
### From build folder

``` bash
cd build
phing
```
### From build folder using bash script

``` bash
cd build
./build.sh
```
### From extension root

You have to duplicate `build/build.properties.dist` to the extension root `build.properties` and then adjust the content to your test site path like:

``` ini
# Properties files are configuration files used by PHING scripts containing:
# key=value

# Testing Joomla site localhost folder used by build.xml
joomla.dir=/var/www/jcms3x
```

Then you can update the test site anytime with:

``` bash
phing -f build/copier.xml
```
### From build folder

You have to duplicate `build/build.properties.dist` to `build/build.properties` and then adjust the content to your test site path like:

``` ini
# Properties files are configuration files used by PHING scripts containing:
# key=value

# Testing Joomla site localhost folder used by build.xml
joomla.dir=/var/www/jcms3x
```

Then you can update the test site anytime with:

``` bash
cd build
phing -f copier.xml
```

PR #184, opened 2014-07-15
Gitkeep packaging and packages folders

javigomez:
This pull should solve the issue raised by JM here: https://github.com/joomla-projects/com_localise/pull/182#issuecomment-48993351

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,9 +35,9 @@ build.properties
 *.tar.*
 
 # Packages directories #
-build/packages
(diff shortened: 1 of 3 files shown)


phproberto:
As commented on the Skype chat:

"is better to have the folder structure always created Javier so people knows what are we talking about when we comment about that folders"

So I'd rebase it and submit

javigomez:
@phproberto rebased ^_^

PR #187, opened 2014-07-15
Missing weekEnd metadata in core language xmls

infograf768:
This adds in the xx-XX.xml the weekend metadata field.

```
<weekEnd>0,6</weekEnd>
```


javigomez:
Works, but @infograf768  can you improve the description?

screen shot 2014-07-17 at 12 07 43

I think is not easy to understand what to put there with the current description

PR #188, opened 2014-07-25
Add travis PHP Parse Error checker

javigomez:
This parse error checker is a PHP script that checks the existence of PHP parse Errors in Travis:
https://github.com/redCOMPONENT-COM/redITEM2/blob/develop/.travis/phpparseerrorchecker.php

Initially we used 

``` sh
- find ./component -name '*.php' -exec php -l {} \;
```

But it generates an endless log in travis that is not very much useful:

``` sh
$ find ./component -name '*.php' -exec php -l {} \;
No syntax errors detected in ./administrator/component//views/categories/tmpl/elements.php
No syntax errors detected in ./component/admin/views/categories/tmpl/default.php
No syntax errors detected in ./component/admin/views/categories/view.html.php
...
```

After that we thought on:

``` sh
- find ./component -name '*.php' -exec php -l {} \; | grep "Parse error";
```

But GREP exits with **exit(0)** if founds a "parse error" and with **exit(1)** if it doesn't found, witch is exactly the opposite that we need to make Travis fail when there is a Parse error.

So the proposal is /.travis/phppec.php takes control of the exit code and makes Travis fail if there is a Parse error found.

@mbabker, @phproberto, @wilsonge If you like this proposal we can add it too to the Joomla-cms repository

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,6 @@ before_script:
   - composer install
 
 script:
-  - find ./component -name '*.php' -exec php -l {} \;
+  - php .travis/phppec.php administrator/
   - php .travis/phpcs.php
   - php .travis/phpmd.php
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,6 @@ before_script:
   - composer install
 
 script:
-  - php .travis/phppec.php administrator/
+  - php .travis/phppec.php component/
   - php .travis/phpcs.php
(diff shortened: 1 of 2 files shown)


javigomez:
See it in action:
screen shot 2014-07-25 at 15 10 09

Or when there is an error:
screen shot 2014-07-25 at 15 11 49

wilsonge:
I like it :)

PR #190, opened 2014-08-13
Proposing core translations only when a xx-XX.xml file is present in the xx-XX folder

infograf768:
Make sure you have a xx-XX (or any tag of a language not installed or for which no xml has yet been created in a core folder (administrator should contain some of them as com_localise copies the languages of its folder into admin lang package).
Filter in the languages view by this language (I used here da-DK) and by administrator, make sure list is set to 20 only (because of the memory limit issue): no language should show.
Then swicth to the translations view:
It will propose all core files for xx-XX (da-DK in my test case) to translate.
This is very wrong.

Go back to languages  and Purge database.
Patch with this PR.
Switch again to translations.

Test for an existing installed language AND for a language for which you have created an xml in the language view. All should be fine.

Because of code style changes (indentation), this PR looks complex, but it just checks if an xx-XX.xml exists each time we have a "foreach ($tags as $tag)"

```
                    foreach ($tags as $tag)
                    {
                        if (JFile::exists($path . '/' . $tag . '/' . $tag . '.xml'))
                        {
```

or

```
if (JFile::exists($client_folder . '/' . $tag . '/' . $tag . '.xml'))
```

### DIFF
--- a/component/admin/models/forms/translations.xml
+++ b/component/admin/models/forms/translations.xml
@@ -48,6 +48,9 @@
 				onchange="this.form.submit()">
 				<option value="">COM_LOCALISE_OPTION_TRANSLATIONS_TYPE_SELECT</option>
 			</field>
+			<field name="spacer_client" type="spacer" class="text"
+				label="COM_LOCALISE_OPTION_CLIENT_SELECT"
+			/>
 			<field
 				id="client"
 				name="client"
@@ -55,8 +58,10 @@
 				class="span12 small"
 				hidden="true"
 				onchange="this.form.submit()">
-				<option value="">COM_LOCALISE_OPTION_CLIENT_SELECT</option>
 			</field>
+			<field name="spacer_lang" type="spacer" class="text"
+				label="COM_LOCALISE_OPTION_LANGUAGE_SELECT"
+			/>
 			<field
 				id="tag"
 				name="tag"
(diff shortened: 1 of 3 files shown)


infograf768:
With this last commit, we solve the memory limit by forcing the translations view to cope with only one client and one language at a time.

If one switches from the languages view where the filters would be set to Select client and Select language (i.e. all languages and all clients) to the Translations view, then the translations view will be forced into Site client and Reference language display, thus needing to set the filter as desired not creating anymore memory issues.

Languages view:
screen shot 2014-08-15 at 09 58 50_languages

Translations view

screen shot 2014-08-15 at 10 19 53

remains, as can be seen, to display the right client and language in the filters fields

javigomez:
Tested. works!

@infograf768 what do you think if we change the way we navigate through the component. 

The proposal is to remove the left menu.

Then the only way to get to the "translations" view is to click through the "files" link:

screen shot 2014-08-15 at 11 25 44

That means that you are forced to filter. Would it have sense?

infograf768:
Javier, I do not think this was fully ready to merge: we do not get the right value in the filter fields:
I.e we can't change the client AND language 

If I change language, I get the correct results for client Site and the new language (example: fr-FR) < but the filter now stays on Afrikaans (first lang in the list) and I never can get Administrator and fr-FR

Concerning your proposal to use the icon from the languages view, this would mean that we would have to go back to languages for any new translations display: language or client

infograf768:
https://github.com/joomla-projects/com_localise/issues/194 should now solve the filters issue.

PR #194, opened 2014-08-15
Proposing core translations only when a xx-XX.xml file is present in the xx-XX folder + using one language and one client only in translations view to cope with memory issues

infograf768:
replaces https://github.com/joomla-projects/com_localise/pull/190 which was merged per error and then reverted

See there for instructions.
# DO NOT MERGE until the filters issue is fixed

https://github.com/joomla-projects/com_localise/pull/190#issuecomment-52289516

### DIFF
--- a/component/admin/language/en-GB/en-GB.com_localise.ini
+++ b/component/admin/language/en-GB/en-GB.com_localise.ini
@@ -206,6 +206,7 @@ COM_LOCALISE_TOOLTIP_TRANSLATION_AZURE="<b>Windows Azure</b><br/>Click to find a
 
 [Translations]
 
+COM_LOCALISE_ERROR_CHOOSE_LANG_CLIENT="Please choose a client AND a language in the filters before switching to the Translations view"
 COM_LOCALISE_HEADER_TRANSLATIONS="Translations"
 COM_LOCALISE_HEADING_TRANSLATIONS_AUTHOR="Author"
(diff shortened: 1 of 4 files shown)


infograf768:
We will now get an error if a client and a language are not chosen in the filters when switching to Translations view:

screen shot 2014-08-18 at 11 21 59

fastslack:
tested works very nice when Joomla! debug is OFF. Joomla! debug on return:

```
Fatal error: Allowed memory size of 268435456 bytes exhausted
```

Seems that Joomla debug has some issues, its not related with this patch

Achal-Aggarwal:
@test
Works seamlessly in both debug modes.

Suggestion: If we are not going to remove translations link from left menu (which I think is correct) can we remove it from upper main drop down menu. Since it will just redirects to Languages view.

Or let the user open translations with no client & language selected by default (or from Languages view) and give a blank view with filters "Client" & "Language" at the top.

infograf768:
```
Suggestion: If we are not going to remove translations link from left menu (which I think is correct) can we remove it from upper main drop down menu. Since it will just redirects to Languages view.
```

In fact, one can switch to another component and then come back directly to the Translations view which keeps its previous settings for the client and language if it has been set before. So the menu =>submenu can be used.

```
Or let the user open translations with no client & language selected by default (or from Languages view) and give a blank view with filters "Client" & "Language" at the top.
```

I did not succeed coding this.

As this works, let me merge it and we will refactor/improve in the future if possible.

PR #196, opened 2014-08-20
Include media custom calendar in core pack when available

infograf768:
To test, install fa-IR (Persian).
Then create a new core package using fa-IR as language in the packages view
Download the package (for example: fa-IR_joomla_lang_full_3.3.3v2.zip)
screen shot 2014-08-20 at 13 05 44

Uncompress the pack and then the site_fa-IR.zip:
The package will now include the necessary js folder in the site folder and the site install.xml will contain

```
    <media destination="fa-IR">
        <filename>index.html</filename>
        <filename>js/calendar-setup.js</filename>
        <filename>js/index.html</filename>
        <filename>js/calendar.js</filename>
    </media>
```

NOTE: remains to add the possibility in the package view to upload such a fa-IR folder in the media folder if not present (if feasible).


Achal-Aggarwal:
@test Did all the steps as mentioned and got the result. Test is passing.

infograf768:
Thanks. Merging.

PR #197, opened 2014-08-20
Translations

Achal-Aggarwal:
It is based on #194. It looks nice if we don't redirect the user back to languages if client and/or language is not set and rather display a notice in translations view only. Through this PR we can do that.


infograf768:
This works fine now. Thanks!

Issue #198, opened 2014-08-21
Saving a new language behaves as Save and New


infograf768:
in the languages view, create a new language and save (not save and close).
The new language is indeed created but the redirect displays an empty form

Achal-Aggarwal:
It is cms issue, here is the patch
https://github.com/Achal-Aggarwal/joomla-cms/compare/joomla:staging...Achal-Aggarwal:fix-com_local-198.diff
If it solves the problem here then I'll send a PR.

PR #200, opened 2014-08-22
Solving issues with new language save

infograf768:
This is a common work with @Achal-Aggarwal.

It solves https://github.com/joomla-projects/com_localise/issues/198
It also prevents the new saved language to have the wrong client displayed on redirect after Save. 
Also it makes the client field readonly preventing issues when an edited language xml already exists for that client.

To test: create new language xml for different clients after setting the Languages view settings to different values for client, including all clients. Save (Not Save and Close) and check that the xml is the correct one displayed.

### DIFF
--- a/component/admin/models/language.php
+++ b/component/admin/models/language.php
@@ -103,6 +103,12 @@ class LocaliseModelLanguage extends JModelForm
 		// Get the form.
 		$form = $this->loadForm('com_localise.language', 'language', array('control'   => 'jform', 'load_data' => $loadData));
 
+		// Make Client field readonly when the file exists
+		if ($this->getState('language.id'))
+		{
+			$form->setFieldAttribute('client', 'readonly', 'true');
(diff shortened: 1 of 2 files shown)


component/admin/models/language.php
```diff
@@ -371,6 +377,10 @@ public function save($data = array())
 			}
 
 			$id = LocaliseHelper::getFileId($path);
+
+			// Dummy call to populate state
```
phproberto, line 381: Why not call directly `$this->populateState()` then?

Achal-Aggarwal, line 381: Yeah, we can do that also. But actually our aim is to not populate state after setState (called at L384).
we are setting language.id in this save method and in its parent save method a getState is called upon 'language.id' which again populates state. The only flag to stop populating state is $this->__state_set which is only modified in getState after calling populateState.

Issue #17, opened 2014-04-11
Models: Use getUserStateFromRequest on populateState


phproberto:
In some models I saw that component is wrongly populating model state. Example:

https://github.com/joomla-projects/com_localise/blob/develop/administrator/components/com_localise/models/downloadpackage.php#L29-L30

To populate the state we should be using something like this:

``` php
$name = $this->getUserStateFromRequest($this->_context . '.name', 'name');
```

Then we can probably remove the `getForm()` method that is double checking the data in session.

javigomez:
assigning to @gunjanpatel

gunjanpatel:
Updating link https://github.com/joomla-projects/com_localise/blob/develop/component/admin/models/downloadpackage.php#L30-L31

gunjanpatel:
@phproberto and @javigomez sent fix for one view. Please review.

infograf768:
https://github.com/gunjanpatel/com_localise/commit/031334993f1ea82eb8c5dc910869519178c3f869
This does not work: we can' create a new language.

```
[24-Aug-2014 16:09:51 UTC] PHP Notice:  Undefined index: client in /Applications/MAMP/htdocs/trunkgitnew/administrator/components/com_localise/controllers/language.php on line 82
[24-Aug-2014 16:09:51 UTC] PHP Stack trace:
[24-Aug-2014 16:09:51 UTC] PHP   1. {main}() /Applications/MAMP/htdocs/trunkgitnew/administrator/index.php:0
[24-Aug-2014 16:09:51 UTC] PHP   2. JApplicationCms->execute() /Applications/MAMP/htdocs/trunkgitnew/administrator/index.php:42
[24-Aug-2014 16:09:51 UTC] PHP   3. JApplicationAdministrator->doExecute() /Applications/MAMP/htdocs/trunkgitnew/libraries/cms/application/cms.php:251
[24-Aug-2014 16:09:51 UTC] PHP   4. JApplicationAdministrator->dispatch() /Applications/MAMP/htdocs/trunkgitnew/libraries/cms/application/administrator.php:152
[24-Aug-2014 16:09:51 UTC] PHP   5. JComponentHelper::renderComponent() /Applications/MAMP/htdocs/trunkgitnew/libraries/cms/application/administrator.php:98
[24-Aug-2014 16:09:51 UTC] PHP   6. JComponentHelper::executeComponent() /Applications/MAMP/htdocs/trunkgitnew/libraries/cms/component/helper.php:332
[...]

PR #202, opened 2014-08-24
[review #17] Models: Use getUserStateFromRequest on populateState

gunjanpatel:
Please review the changes for this one view, then we may proceed for all.

Here class is extended from `JModelForm` https://github.com/joomla-projects/com_localise/blob/develop/component/admin/models/language.php#L21 that's why we can use,

``` Php
$name = $this->getUserStateFromRequest($this->_context . '.name', 'name');
```

**Contains fix for #17**

_Question: We should use `JModelAdmin` here or it's fine with `JModelForm`?_

### DIFF
--- a/component/admin/models/language.php
+++ b/component/admin/models/language.php
@@ -20,6 +20,11 @@ jimport('joomla.access.rules');
  */
 class LocaliseModelLanguage extends JModelForm
 {
+	/**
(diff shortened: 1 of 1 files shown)


javigomez:
tested!  I'm not having issue. merging.

Issue #208, opened 2014-08-28
Checkin-checkout broken.


infograf768:
It is broken for all : language, package, translation.

PR #211, opened 2014-08-30
Displaying the checkout icon when necessary

infograf768:
https://github.com/joomla-projects/com_localise/issues/208
To test use 2 browsers and edit a package, a translation or a language.
before patch = errors
after patch: the icon displays, example below for packages

screen shot 2014-08-30 at 19 35 35

IMPORTANT: this is incomplete!
Instead of redirecting to the manager when someone clicks on the link of a checkedout item, it opens the edit page and as the close does not work, the only way for the user is to get back in browser.

### DIFF
--- a/component/admin/controllers/languages.php
+++ b/component/admin/controllers/languages.php
@@ -16,7 +16,7 @@ defined('_JEXEC') or die;
  * @subpackage  Localise
  * @since       1.0
  */
-class LocaliseControllerLanguages extends JControllerLegacy
+class LocaliseControllerLanguages extends JControllerAdmin
 {
 	/**
 	 * Method to purge the localise table.
--- a/component/admin/controllers/packages.php
+++ b/component/admin/controllers/packages.php
@@ -16,7 +16,7 @@ defined('_JEXEC') or die;
  * @subpackage  Localise
  * @since       1.0
  */
-class LocaliseControllerPackages extends JControllerLegacy
+class LocaliseControllerPackages extends JControllerAdmin
 {
 	/**
 	 * Proxy for getModel.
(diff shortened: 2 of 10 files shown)


mbabker:
There's 2 issues that will keep this from working:

1) None of your controllers are extending `JControllerAdmin` where the main checkin function exists

2) The param before `$canCheckin` in your `JHtml` call should be the subcontroller you're calling.  Using `component/admin/views/languages/tmpl/default_body.php` as an example, the line should read this:

``` php
<?php echo JHtml::_('jgrid.checkedout', $item->editor, $item->checked_out_time, 'languages.', $canCheckin); ?>
```

That would call `LocaliseControllerLanguages` which should extend `JControllerAdmin`.

mbabker:
Just discovered we're also missing a param in the `JHtml` statement.  Should actually be this:

``` php
<?php echo JHtml::_('jgrid.checkedout', $i, $item->editor, $item->checked_out_time, 'languages.', $canCheckin); ?>
```

infograf768:
Changed the languages controller to extend JControllerAdmin
But same issue

infograf768:
The only way I found to get the error message in the manager and not the edit screen was to override the checkin/checkout methods in the models. This worked also for packages.
I could not get the icon to checkin an item when user is authorised to checkin via com_checkin but such a user can edit the item which checks it in de facto.

To test:
One browser should have a user logged with com_checkin permissions, the other browser one without.

infograf768:
Closing this one as https://github.com/joomla-projects/com_localise/pull/212 solved the issue for languages and translations. 
Remains to do it for packages.

PR #213, opened 2014-09-09
Fixes checkin-checkout behavior for Packages.

Achal-Aggarwal:


infograf768:
Merged. This solves OK the checkin-checkout.
Achal and I found 3 more aspects to deal with n packages:
1. Save a package (new or duplicated) with the name of an existing one replaces the existing one. Needs an alert here
2. When deleting a package in the packages view or through $oldpath, the db is not updated, whether in the _localise or _assets table.
3. We need to thrown an error and a redirect if a package to edit is no more in the file system

Achal-Aggarwal:
PR for issue one : https://github.com/joomla-projects/com_localise/pull/214

PR #214, opened 2014-09-11
Check existing package on creation of a new package.

Achal-Aggarwal:


infograf768:
@Achal-Aggarwal
This works only when creating a new package, not when changing the name of a package (for example a duplicated one).
Would not it be more friendly to just check if a pack exists with the same name and throw a js confirmation Alert asking the user if he wants to overwrite the existing pack or not?

Achal-Aggarwal:
Yeah it would be better but only if we do it when saving an existing package. While creating a new package with same name of any existing one, throwing an error is more appropriate.

infograf768:
Let's commit 216 first and then this one with the correct new string

infograf768:
I have posted https://github.com/joomla-projects/com_localise/pull/219

Maybe that would be enough?

infograf768:
I merged #219 in the mean while as we have to move on

Achal-Aggarwal:
@infograf768  How about we do the same thing as we have in com_content.
Ex:- While creating a new article if its alias matches any existing then it just throws an error. And on update also it does the same. More easy and less complex.

infograf768:
Why not. If it works :)

infograf768:
obsolete

PR #215, opened 2014-09-17
[fix] strings containing HTML code are not saved

phproberto:
## Reproduce the issue

Edit a translation with standard editor (clicking on the name). Modify any string to add HTML code inside. Example: "My language string for <strong>Jean-Marie</strong>" (note the strong tags)
## Expected result

HTML is saved
## Result

HTML is cleared because the way that $_POST data is received by the model
## Patch test

Apply the patch and check it. HTML should be saved correctly.


infograf768:
Great! works fine

Achal-Aggarwal:
@test works for me. Thanks.

PR #216, opened 2014-09-17
Deleting db rows when deleting packages

infograf768:
Test: Duplicate some packages in the packages manager.
Tick some of the duplicated packages and Click on Delete button.

Now the rows in the _assets table as well as in the _localise table will also be deleted.


infograf768:
Added code to delete original rows in in the _assets table and in the _localise table when modifying a package name.
To test, duplicate a core package or an extension package, change its name and save.

infograf768:
While we wait for a cleaner solution, I merge this as it works

PR #218, opened 2014-09-25
Update en-GB.com_localise.ini

brianteeman:
In Joomla we have standardised on writing checked in and not checked-in


infograf768:
thanks

PR #219, opened 2014-09-25
Displaying a confirm alert when saving a package

infograf768:
To test, save any package.


infograf768:
Merging to be able to create a new release and solve some other issues. 
Can be reconsidered if new proposal

PR #221, opened 2014-10-01
Adding core340.xml to deal with coming release. Letting customise this reference in configuration

infograf768:
The core(xxx).xml is used 
1. To add the correct core icon in the Translations manager
2. when creating extensions packages in order to not propose core files to the selection.
To test, install the French pack which contains geshi.
Then edit media/com_localise/packages/core340.xml and add 
`<filename>plg_geshi</filename>' in the administrator group.
before patch: the file will be available when creating an extension package
after patch, it will not show anymore in the list


infograf768:
Will create asap a new master core

infograf768:
This does not work for translations. Core icon does not display in Translations Manager;

Also found out that (without this PR) even when updating the core.xml file to 3.4.0 state (4 plugin files more) and adding these files in the en-GB administrator folder, I could not get the new files listed to be translated, although they show when displaying en-GB

infograf768:
closing this. We just need to update core.xml and add a master_core_340.xml to be duplicated.
If the user tries to download a final pack based on a duplicate of master_core_340.xml from a site which does not contain the new lang files of 340, they will just not be included and no error would show as they would not be in reference

PR #222, opened 2014-10-03
Displaying in Translations Manager the ini files from plugins' language folder

infograf768:
The plugin language files when only available in the plugin folder are not proposed to translation.
This PR will display them.
To test, install https://www.dropbox.com/s/c9o8sxtud7swebj/jostag_plugin.zip?dl=0
Then choose in Translations Manager an installed language (not fr-FR or en-GB) 
test before and after patch

This does not totally solve the issue as we can translate the .ini but not the .sys.ini.
It seems that one is not loaded.

### DIFF
--- a/component/admin/helpers/localise.php
+++ b/component/admin/helpers/localise.php
@@ -572,8 +572,8 @@ abstract class LocaliseHelper
 					$parts  = explode('_', $extension);
 					$group  = $parts[1];
 					$parts	= explode('.', $filename);
-					$filename = $parts[0];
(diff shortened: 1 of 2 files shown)


infograf768:
The plugin case did not cope with .sys.ini. Corrected helper

infograf768:
This sitll needs to be looked at before merging as the 3rd party plugins do also display in "Site" after patch.

infograf768:
Corrected to prevent plugins to be displayed when filtering by site

PR #227, opened 2014-10-09
updated ja-JP.

Richell:


javigomez:
Thanks @Richell  :) 

My japanese is not good enough for do a review. @infograf768 all yours :)

PR #230, opened 2014-10-10
Bug when using Import Specific Files

infograf768:
Test: 
display Packages manager
Edit a core package for a language (not master).
Click on Import Specific Files
(you will have beforehand created somewhere a file called xx-XX.css)
Browse to pick this file and Import.

We get an error as $name is not found, therefore the xml can't be read.
Patch and test again. This time we are using the $id.


wilsonge:
Before patch get error

```
To be able to import a specific file, the core package has first to be saved.

Error importing the file
```

After test file uploaded successfully.

wilsonge:
Merged. Thanks JM.

PR #233, opened 2014-11-08
Fixed left sidebar layout issue.

Achal-Aggarwal:
Before:
image

After:
image
image


infograf768:
Please explain what are the results of the changes~:
screenshot before and after.
For me here it breaks totally the display
when collapsed, the table disapears
when not collapsed, the table displays below the sidebar

infograf768:
The reason we have these 2 closing divs for the moment is that the filters are not added to the sidebar.
Therefore was added a custom layout /sidebars/submenu where they are absent.
Once we have modified the code to be able to use JHtmlSidebar::addFilter(), we will be able to correct this.

Achal-Aggarwal:
We need to take care of this once https://github.com/joomla/joomla-cms/pull/4450 is sorted out.

infograf768:
obsolete

PR #240, opened 2014-11-08
Making method signature same as its parent and adding doc comment for pa...

Achal-Aggarwal:
...rams.


wilsonge:
:+1:

infograf768:
let's commit this after #245

Issue #241, opened 2014-11-08
Autofill the creation date field when creating new language


BrittaStender:
![creation-date-bug](https://cloud.githubusercontent.com/assets/2777672/4965870/487c5c78-679b-11e4-9073-9025c010861d.png)

It states "now" as standard text. Couldn't it contain the actual date instead?

wilsonge:
Fix in #243

infograf768:
Look at discussion in https://github.com/joomla-projects/com_localise/pull/243

PR #243, opened 2014-11-08
Type calendar field for creation date

### LINKED ISSUE #232: Date notification "now" in form "Create new Language"
Components > Localise > Languages > Create new language
The Creation Date has "now" as default field,
after saving, the date field stores "now" as date.

In my example file /administrator/language/ex-EX/ex-EX.xml
the date is hardcoded stored as "now":
<creationDate>now</creationDate>

I think it would be better to display the current system date instead.

wilsonge:
Fixes #232 and #241

### DIFF
--- a/component/admin/models/forms/language.xml
+++ b/component/admin/models/forms/language.xml
@@ -120,7 +120,7 @@
 			description="COM_LOCALISE_LABEL_LANGUAGE_VERSION_DESC" />
 		<field
 			name="creationDate"
-			type="text"
+			type="calendar"
 			default="now"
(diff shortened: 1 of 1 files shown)


infograf768:
The problem here is that some TTs have not/do not normalise(d) to year-month-date in English for their language xx-XX.xml. This is why we took off the field as calendar at the beginning of the process here as we got an error.
I do agree though that "now" is not a good default.
maybe shall we just not enter any default

pe7er:
I prefer "year-month-date" but if it will cause difficulties, then an empty field might be a better choice

infograf768:
@pe7er 
Please see https://github.com/joomla-projects/com_localise/pull/245

PR #245, opened 2014-11-10
Improving checks when creating a language xml

infograf768:
To test, create a new language
/index.php?option=com_localise&view=language&layout=edit&client=site&tag=
This patch makes some fields required:
Creation date
Version
Locales

It also takes off the default="now" for the creation date. ( https://github.com/joomla-projects/com_localise/pull/243 )

It also displays error messages when the tag entered is not following the correct format
xx-XX or xxx-XX

I.e. this will trigger an error
xxxx-XX
xxX-XX
xx-xx
or combined.

It also triggers an error is the language name if not entered to replace the proposed default format which is just a tip


wilsonge:
Hey JM Just seen this when I got back to england from JWC. Looks acceptable to me :)

Issue #238, opened 2014-11-08
Packages overview shows very little information


pe7er:
See Component > Localise > Packages
The list of packages only show: Title + Type

Is it possible to add somewhat more information about the packages?
I am thinking about: language tag, location (admin/site) and creation date (e.g file date of xml/zip file) ?

infograf768:
This manager only displays the internal to com_localise packages, not the .zip packages.
Therefore, we can't show the date which is created when the .zip is created.
The location is both site/admin for core packages, it can be one, the other, or both for a 3rd party extension.
Looking how to get the Tag and version now

infograf768:
please test
https://github.com/joomla-projects/com_localise/pull/247

PR #247, opened 2014-11-11
Improving packages Manager columns and filtering

infograf768:
See https://github.com/joomla-projects/com_localise/issues/238
after patch you will get:

screen shot 2014-11-11 at 11 10 10


Achal-Aggarwal:
@test :+1:

PR #248, opened 2014-11-13
Make locale field disabled and not required if installation is selected ...

Achal-Aggarwal:
...as client.


infograf768:
Thanks Achal and Roberto! Merging.

PR #249, opened 2014-11-13
Delete/rebuild database when deleting a language

infograf768:
When we delete a language xx-XX.xml for s specific client in the languages manager and have created some translations for that language, these remain in the database, cluttering it as it is populated with all ini paths for that language and client.
This PR will purge the database when deleting a language, forcing the database to be repopulated in the managers.

To test, create an xx-XX.xml for site. Then save some translations for that language (no need to translate. In fact just displaying the translations manager for the language/client will create the rows in the db). Look the rows in the _localise table.
delete the language. Reload and look at the rows in the db.

### DIFF
--- a/component/admin/helpers/localise.php
+++ b/component/admin/helpers/localise.php
@@ -747,4 +747,58 @@ abstract class LocaliseHelper
 
 		return $sections[$filename];
 	}
+
+	/**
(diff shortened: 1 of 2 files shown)


phproberto:
Please check https://github.com/joomla-projects/com_localise/pull/250

PR #250, opened 2014-11-13
[fix] delete language translations when deleting a language

phproberto:
This tries to implement https://github.com/joomla-projects/com_localise/pull/249 in a better way:
- Not using a helper to interact with DB. Table related stuff shoudl be inside table.
- Avoid to fully purge DB and then to regenerate files.

DB structure is not ok anyway. I think we should consider to add a language tag column to `#__localise` so everything would be easier.


infograf768:
Works great. Merging, Closing #249 

Note: we may indeed need more columns: lang tag, client, type of file (ini or xml)
This would mean a big refactoring.

PR #255, opened 2015-01-12
the previous link was invalid

kshitijSharma2014:


kshitijSharma2014:
@puneet0191 
I don't know if i am right but the previous mentioned link was giving me a 404 error.
also I checked if there was any repo names redshop and there wasn't.

puneet0191:
@kshitijSharma2014 Thanks for catching this, I forgot to update the comment, I took this file from a private repo.

javigomez:
Actually the right address is: https://github.com/joomla-projects/com_localise/blob/develop/tests/acceptance.suite.dist.yml

PR #256, opened 2015-01-24
Com Localise with special keys cases

Valc:
I have forked the com_localise to show few features than maybe can to be added with improved source at the main branch. 
Here is explained with screenshots what is new:
https://github.com/Valc/com_localise/releases/tag/4.0.8-dev_by_valc

Regards, Valc

PR #258, opened 2015-01-26
Correcting format and strings in #256

infograf768:
This replaces #256.
It should have much less format errors. I also corrected some language strings, differentiating between Keys and strings.
Let's see what Travis says.


infograf768:
Closing as it is integrated in https://github.com/joomla-projects/com_localise/pull/259

PR #261, opened 2015-01-27
Fix icon in the toolbar for the language view

zero-24:
This PR fix the icon in the toolbar for the language view with current staging

bevor
bevor_patch

after
after_patch


zero-24:
This fix now two more places one fallback to the default the other to the comments.

Icon set: http://www.nonr.nl/jui/index.php/icons

infograf768:
Good find!. Thanks.
Waiting for https://github.com/joomla-projects/com_localise/pull/260 to be solved and I will merge.

infograf768:
As #260 will take some more time, merging.

zero-24:
Thanks @infograf768

PR #266, opened 2015-02-07
Choosing reference file priority

infograf768:
To test before patch:
In core, change manually some words in the value of a string for an extension which has his language files at the same time in the extension folder and in the core folder.
A simple example is the languagecode system plugin.
Change for example in the file in the core folder (en-GB.plg_system_languagecode.ini) the string
`PLG_SYSTEM_LANGUAGECODE_FIELDSET_LABEL="Language codes"`
to
`PLG_SYSTEM_LANGUAGECODE_FIELDSET_LABEL="MY language codes"`

then load com_localise with en-GB as reference, another language as target, and edit the file.
You will still see
screen shot 2015-02-07 at 18 47 24

This is because com_localise gives priority to the "local" file (i.e. the file in the extension/language/en-GB/folder instead of the file in the core folder ("global").

Now patch and test:
If you do not touch at the Options, the priority is now to the "global" folder.
Editing the file will display the correct value above which you changed in the core folder.

Then play with the new parameter in the Options (Default is No):
"Priority to extension folder"

You will get the correct value depending on these settings.

### DIFF
--- a/component/admin/config.xml
+++ b/component/admin/config.xml
@@ -72,6 +72,16 @@
 			label="COM_LOCALISE_LABEL_SUFFIXES"
 			description="COM_LOCALISE_LABEL_SUFFIXES_DESC" />
 		<field
+			name="priority"
+			type="radio"
+			class="btn-group btn-group-yesno"
+			default="0"
+			label="COM_LOCALISE_LABEL_PRIORITY"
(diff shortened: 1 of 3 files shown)


Valc:
After apply is working fine here.
Regards.

PR #268, opened 2015-02-08
Keys filters

Valc:
Following a JM idea, with this pull now we can to add in the "Localisation Manager: Edit translation" => "Strings" tab page a filter to filter by keys status (unchanged, translated, untranslated, etc.).

Seems is working and maybe can to be added in main brach with improved code.

Here there is  a sample with untranslated strings:
untranslated_filter

Regards.


infograf768:
Can you update as I merged 2 PRs?

infograf768:
In fact just take out the changes to models/translations.php

This works real good here

Valc:
Done :D

infograf768:
@valc
we already have these strings:
COM_LOCALISE_TEXT_TRANSLATION_TRANSLATED="Translated"
COM_LOCALISE_TEXT_TRANSLATION_UNCHANGED="Unchanged"
COM_LOCALISE_TEXT_TRANSLATION_UNTRANSLATED="Untranslated"

and
JALL="All"
JOPTION_SELECT_PUBLISHED="- Select Status -"

No use to create new strings for the filters.

infograf768:
Did a bit of cleaning. 

Also I see no reason for sprintf in the filter Options, example:
`JText::sprintf('COM_LOCALISE_FILTER_ALL_KEYS')`

infograf768:
should be OK now. :)

Valc:
Yep, i had not noticed about that: well done.
Here is working great: happy to catch untranslated strings within large files  in record times :D.

PR #269, opened 2015-02-09
Listing available languages in the language filters depending on the client settings

infograf768:
Before this PR, all available languages, for whatever client they do really exist (i.e. where there is an xx-XX.xml), will be displayed as a choice in the language filters in Languages Manager and Translations Manager,
This PR will only display the available language depending on the client.
Before patch (example client is site but all installation-only languages display:

screen shot 2015-02-09 at 12 20 35

After patch, we will get:

screen shot 2015-02-09 at 12 18 00

### DIFF
--- a/component/admin/models/fields/language.php
+++ b/component/admin/models/fields/language.php
@@ -37,6 +37,8 @@ class JFormFieldLanguage extends JFormFieldList
 	 */
 	protected function getOptions()
 	{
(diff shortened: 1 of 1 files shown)


Valc:
I was testing 10 minutes ago and is working as is explained

infograf768:
:)

I will now release 4.0.11-dev

Issue #254, opened 2014-12-12
Crate a bridge with translation services


javigomez:
This is an experimental feature but it would be awesome that com_localise will be able to have an API that allows it to send and retrieve translations from services like Transifex or Crowdin.

Adding here so we don't forget what we spoke during JWC14 @yireo

infograf768:
Transifex is not good enough for J core as it does not take care of supplementary strings depending on languages (plurals, others)

javigomez:
Yes, Transifex can't do this: https://github.com/joomla-projects/translate-joomla#working-with-plurals

Still having an API will allow us to export and import files and do crazy things. Having a webservices layer for com_localise would allow us to simplify tasks like https://github.com/joomla-projects/com_localise/issues/57 or for example to connect it with Slack, so any new released language package sends a notification to Slack.

This feature is not necessarily a bridge with transifex but an experimentation with opening the tool for interacting with other tools.

fastslack:
This can be done using some ajax.raw.php and returning the required data via JSON. What kind of data we need to export?

javigomez:
Well, is not clear yet. But actually I can think on one translation. That would allow us to do inline edition of translations, or complete files.

But you are right we will need to think it what thinks to we want to allow to be in the api: translation, language file, package...

elkuku:
> Transifex is not good enough for J core as it does not take care of supplementary strings depending on languages (plurals, others)

With "others" you probably mean the special `QQ` constant?
As this is about to disappear (hopefully), Transifex maybe now "good enough" for our needs?
Not sure that Transifex really supports the special plural string handling but hey...

@javigomez this is an awesome idea :wink:

infograf768:
> With "others" you probably mean the special QQ constant?

Nope. I mean specific language strings as (in fr-FR):

```
MOD_MENU_HELP_SUPPORT_CUSTOM_FORUM="Forum joomla.fr"
MOD_MENU_HELP_SUPPORT_OFFICIAL_LANGUAGE_FORUM="Forum officiel français"
MOD_MENU_HELP_SUPPORT_OFFICIAL_LANGUAGE_FORUM_VALUE="19"
```

or

```
INSTL_SAMPLE_BLOG_FR_SET="Données exemples de type blog, en français"
INSTL_SAMPLE_BROCHURE_FR_SET="Données exemples de type brochure, en français"
INSTL_SAMPLE_DATA_FR_SET="Simple page d'accueil, en français"
[...]

PR #270, opened 2015-02-09
[WIP] Remotes

elkuku:
Ref #254

This is an attempt to manage remote repositories for language files.

Currently supported remote types:
- [GitHub](https://github.com)
- [Transifex](https://www.transifex.com)

Example repositories included in this PR:
- **de-DE** from https://github.com/joomlagerman/joomla
- **pt-BR** from https://www.transifex.com/projects/p/j3corelang/

The contents of the repositories will only be fetched, their file names "normalized" and saved to a common file structure.

In a second step (code not included) they can then be compared/replaced with the core files or build the language packages directly from here.
### Info
- http://community.joomla.org/translations/joomla-3-translations.html

See also:
- https://sites.google.com/site/transjoomla/joomla-projects/jcorelang
### Currently this might not work for you because:
1. **Transifex** The `babdev/transifex` library is installed using composer so it is not accessible by the web app (sorry, the CMS ;)).<br />
   **Solution**: Find in `helpers/remotes/Transifex.php` the method `setCredentials()` there is a `require_once()` call that has to be adjusted to your system.
2. **GitHub** The GitHub API implementation included in the CMS has an issue that has been addressed some time ago in newer versions of this library.<br />
   **Solution**: 
   1. Hack the core, find the corresponding package and update the file OR
   2. Update the library (using composer would be nice)

BTW: I did 2.1 so I might describe that in detail if required..
### Testing

After the above is solved you open the remotes view, select the desired remote(s) and click the button `Fetch Remotes`.
A (somewhat ugly) console window should come up, showing the progress of the operation.
When the operation has finished, the files will be in
- `JROOT/administrator/components/com_localise/build/remotes/...`
### Questions
- Is there any name spacing afford currently?
- Can we integrate composer packages?
- How to update the Joomla! GitHub package?

**Thanks!** :wink:


mbabker:
> Is there any name spacing afford currently?

Within this component or the CMS in general?  If the latter, nothing yet.

> Can we integrate composer packages?

Haven't quite worked out what to do for extensions.  One thing I can think of, which is an ugly hack, is to try and include `libraries/composer_autoload.php` to retrieve the Composer ClassLoader object and add the library path via that API.

> How to update the Joomla! GitHub package?

Send PRs to the CMS repo.  If it means we can dump a lot of the duplicated code between the non-namespaced JGithub and the namespaced Joomla\Github, I'd love to see that happen.

elkuku:
> Is there any name spacing afford currently?
> 
> > Within this component or the CMS in general? If the latter, nothing yet.

Actually I meant the former, as the latter would be far out of scope for the PR....

> One thing I can think of, which is an ugly hack, is to try and include libraries/composer_autoload.php

have a look at https://github.com/elkuku/com_localise/blob/remotes/component/admin/helpers/remotes/Transifex.php#L54-55 :wink: 
Maybe we can add a component parameter to set the absolute path to the `vendor` dir?

> Send PRs to the CMS repo

If we could include the composer package, that'd be great :wink:

infograf768:
far over my head. ;)
One question though.

> When the operation has finished, the files will be in
>   JROOT/administrator/components/com_localise/build/remotes/...

and then what, once they are there (and which files exactly) ?

elkuku:
no interest..

PR #273, opened 2015-02-23
Allowing extension packages to be created from the extension language folder files

infograf768:
As of now, a non-core ini file has to be present in the core language folders to be able to create a 3rd party extension package.
A typical pack would show:

screen shot 2015-02-23 at 11 11 08

But we will get an error if the files are only present in the extension language folder.

Depending on the  priority settings (here global is chosen):

screen shot 2015-02-23 at 11 12 04

This patch will create the installable package by first testing if the files exist in the priority-defined place, then in the other. In the case above, it will check first in Global and then Local.


Valc:
I was not able to replicate the issue but now i have noticed that it was due i was using a duplicated base of "master_localise (master_localise) " to get the es-ES package.
EDITED
No, that one is not what i was doing wrong to make it fail: Trying the same duplicating from other bases seems that the com_localise package also is created without issues.
What i need test to make it fail without the patch?
With your patch also is working.

infograf768:
@Valc
You need to make sure the extension ini files for the language tested <b>are ONLY present in the extension language folder</b>, set priority to Global  and test before patch.
You can see where the ini are present (in this case when priority is global) in the translations manager for the language tested by looking at the path:

screen shot 2015-02-24 at 06 46 05

infograf768:
code-wise, it is easy to understand what happens:
we had ONLY if an ini was for administrator
`$path = JPATH_ROOT . '/administrator/language/' . $data['language'] . '/' . $data['language'] . '.' . $translation . '.ini';`
No way to load a $path from the extension language folder.
this is changed in the PR to
`$path = LocaliseHelper::findTranslationPath($client = 'administrator', $tag = $data['language'], $filename = $translation);`

Which, depending on the Priority, looks also in the extension/language folder

Valc:
I had wrongly in mind that com_localise had installed the language files already in extension folder and the issue was there by default, so no more revision for my part trying to ensure the real path and from there all my other wrong steps to test.

Now i can confirm that was failng and after patch is working as is explained.

Regards

infograf768:
Thanks for testing. Now merged.

PR #276, opened 2015-04-23
Strings protection againt max_input_vars limit

Valc:
Hi, :)
Whereas another solution that allows to edit language files without having to send all the strings is implemented, this idea can avoid losing data with large language files edition in this moment.

It is a basic comparation between the the php limit and the total strings in each file but can to be finetuned with the real counter of vars posted by form within language file edition (no idea about how many more are used as default).

Only is showing the idea, it have required Jtext and nice message.

Here the output setting the limit to 100:

captura de pantalla de 2015-04-23 22 43 17

Regards.


Valc:
Another option is handle this one previously to allow the installation of com_localise, with code like this at the 'install.php' file. Here a draft to improve:

```
    public function preflight($type, $parent)
    {
        $max_vars = ini_get('max_input_vars');

        if (version_compare(JVERSION, $this->minCmsVersion, 'lt'))
        {
            JFactory::getApplication()->enqueueMessage(JText::sprintf('COM_LOCALISE_ERROR_INSTALL_JVERSION', $this->minCmsVersion));

            return false;
        }
        elseif ($max_vars > 0 && $max_vars < 10000)
        {
            JFactory::getApplication()->enqueueMessage('It is noticed an issue with big files such as xx-XX.ini case than can make stop the right working mode due your php.ini config.<br />To make it run correctly is required edit your php.ini file and add or change this new limit (by default is set to 1000):<br />max_input_vars = 10000<br />AND ONLY if the environment is running with suhosin also is required add:<br />suhosin.post.max_vars = 10000<br />suhosin.request.max_vars = 10000<br />And then restart apache.<br />');

            return false;
        }

        return true;
    }
```

infograf768:
@Valc

I prefer the original code in the PR as users may not have to deal with very long ini files (for 3rd party extensions for example).

It has one issue though:
$item->total is not accurate as it represents the number of keys, not the number of lines in the file.
Here it looks like it is the latest which counts, or at least also the commented lines.

Also: the item checked is the xx-XX file and not the reference one. I think both should be checked.

infograf768:
Found a way to get the right variables. Will make a new PR when ready.

infograf768:
O have made a new PR here:

https://github.com/joomla-projects/com_localise/pull/287

PR #286, opened 2016-03-16
Support utf8mb4 in com_localise component

andrepereiradasilva:
### Description

This PR is for adding utf8mb4 support to com_localise.

You will need two packages for the tests:
- [package 1] the "official" com_localise 4.0.14-dev release. https://github.com/joomla-projects/com_localise/releases/download/4.0.14-dev/com_localise-4.0.14-dev.zip
- [package 2] A patched com_localise utf8mb4 package that has been created for the tests.
  [com_localise-develop-utf8mb4.zip](https://github.com/joomla-projects/com_localise/files/181332/com_localise-develop-utf8mb4.zip)

This tests need to be made in a mysql db server with (5.5.3+) AND without (lower than 5.5.3) utf8mb4 support.
### How to test

Note: For 3.5.0 tests (WITH or WITHOUT utf8mb4 support) first apply richard patch in joomla latest staging (https://github.com/joomla/joomla-cms/pull/9468).
###### New install
1. Install the patched com_localise utf8mb4 package [package 2]
2. Check in phpmyadmin if `#__localise` and `#__localise_revised_values` tables collations are now **utf8mb4_unicode_ci**
###### Update
1. Install 4.0.14-dev official package [package 1]
2. Check that there is only one table `#__localise` in **utf8_general_ci**
3. Update com_localise with the patched com_localise utf8mb4 package [package 2]
4. Check in phpmyadmin if `#__localise` and `#__localise_revised_values` tables collations are now **utf8mb4_unicode_ci**
##### Joomla 3.4.8

Do exactly the same tests as for Joomla 3.5.0, the only difference will be the results in the Joomla 3.5.0 install and update tests: the result table collations will be **utf8_general_ci**, instead of utf8mb4_unicode_ci.
#### Database server WITHOUT utf8mb4 support (lower than Mysql 5.5.3)

Do exactly the same tests done in `Database server WITH utf8mb4 support (Mysql 5.5.3+)`, the only difference will be the results in the Joomla 3.5.0 install and update tests: the result table collations will be **utf8_general_ci**, instead of utf8mb4_unicode_ci.
### Observations

This test removes also duplicated sql instructions that are already in another PR, but for simplicity they were inserted in this PR too.


andrepereiradasilva:
oh a new PR has merged right now by @infograf768 with new sql tables ... talk about bad luck!

Will have to do a revision to the PR.

richard67:
Well, what I saw looked good to me, same way as the core does it. The DEFAULT='' aded to one of the columns I am not sure if that is necessary, but should not do any harm.

andrepereiradasilva:
> Well, what I saw looked good to me, same way as the core does it. The DEFAULT='' aded to one of the columns I am not sure if that is necessary, but should not do any harm.

Nice. so i will just have to do the same process to the other tables merged in a PR an hour ago ...

richard67:
Yes.

richard67:
@andrepereiradasilva But one thing I should mention: The conversion script has to be run not only when updating to this particular version where utf8mb4 was introduced, it also has to be run when someone in future updates from a version before the one when utf8mb4 was introduced to a version after the one, means as long as you still expect old installations to be converted on update, you have to keep the conversion script in the manifest and run it, and it has to be adapted in future maybe, so it always fits to the data structures of the particular new version.

andrepereiradasilva:
@richard67 i think it's already doing that.
see https://github.com/joomla-projects/com_localise/blob/develop/localise.xml#L28-L33

richard67:
Sure, I just say it has to stay like this also in future version of this component.

The conversion then runs after each update, not doing any harm if already done with last update, but this then covers also the conversion if somone migrated from old utf8 to new utf8mb4 database. Those will be converted to utf8mb4 then as soon as they update this component to a new version, and for the meantime they will still run utf8_unicode_ci (which should not be a problem).

andrepereiradasilva:
yes i think so.

richard67:
The changes in this PR here after last commits still look good to me. So theoretically all is covered.

andrepereiradasilva:
so is the same process that is needed for weblinks too, right?

infograf768:
Folks, I have already prepared a PR, but as it works fine as it is, both on staging and 3.4.8, I think it is not urgent to change.

BTW, this is what it should be as otherwise it breaks (look at  `path`:

```
CREATE TABLE IF NOT EXISTS `#__localise` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `asset_id` int(11) NOT NULL,
  `path` varchar(400) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
  `checked_out` int(10) unsigned NOT NULL DEFAULT '0',
  `checked_out_time` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
  PRIMARY KEY (`id`),
  KEY `idx_path` (`path`(100))
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE=utf8mb4_unicode_ci;

CREATE TABLE IF NOT EXISTS `#__localise_revised_values` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `client` varchar(255) NOT NULL,
  `reftag` varchar(6) NOT NULL,
  `tag` varchar(6) NOT NULL,
  `filename` varchar(255) NOT NULL,
  `revised` tinyint(1) NOT NULL DEFAULT 0,
  `key` varchar(255) NOT NULL,
  `target_text` mediumtext NOT NULL,
  `source_text` mediumtext NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE=utf8mb4_unicode_ci;
```

richard67:
Yes, if they wanna convert + migrate data in this way.

Another was which some more complicated extensions seem to go is to save old data, create new tables and then import back old data.

richard67:
@infograf768 Correct, utf8mb4_bin if path is to be used in join queries or as foreign key.

andrepereiradasilva:
So @infograf768 shall i make that changes (on create and on update) or remove the PR?

infograf768:
Please do and let's talk on Glip tomorrow. :)

andrepereiradasilva:
ok

andrepereiradasilva:
Conflicts fixed, All seems ok now,

infograf768:
@richard67  @andrepereiradasilva 
A question: if we merge this, will com_localise install and work OK on 3.4.8 and below?

andrepereiradasilva:
good question.  i think so.
but tests are needed.

richard67:
Hmm, on a 3.4.8 or previous there is no "query downgrade", i.e. no replacement of "utf8mb4" by "utf8" in sql statements "create table" or "alter table", so it will not work if utf8mb4 is not supported.

richard67:
P.S.: I mean the "query downgrade" of the database driver, JDatabaseDriver::convertUtf8mb4QueryToUtf8().

andrepereiradasilva:
So how can an extension install/upgrade to utf8mb4 and still work also on pre 3.5.0 versions?

richard67:
Was not my idea with the utf8mb4 and how things should be done, I only helped to make it work.

One way of course is to have the extension in 2 variants: 1 for pre 3.5 with sql for utf8 and other one for 3.5.0 and post with utf8mb4.

Or is there a way to sell in manifest files that an sql file shall be run only on certain core versions? I don't think so, but if there is something like this, it could be a way.

Or you leave everythig in utf8 for now and wait until Joomla! has found a strategy for that.

Sorry, it really is a mess, and I did not really think about such things, just seen what was there and was maybe too much focussed on made it work somehow.

infograf768:
I guess we should leave 
com_localise sql as they are for now as it will work on ≦ 3.5.0.

richard67:
Yes, would be the easiest and safest way for now, if you not want to maintain 2 versions, a pre 3.5 ("<", not "≦") and one for 3.5.0 or later.

andrepereiradasilva:
tested the package with utf8mb4 (https://github.com/joomla-projects/com_localise/files/176570/com_localise-develop-utf8mb4.zip) and it installed fine on Joomla 3.4.8

andrepereiradasilva:
so it seems it will "only" affect non utf8mb4 capable db servers, right richard?

richard67:
Sure, on a server supporting utf8mb4 it will work. But not on a mysql not supporting utf8mb4 (server version lower than 5.5.3).

andrepereiradasilva:
thanks, will try also do tests on a 5.1.x server to see exactly what will happen.

infograf768:
Could we devise a script that would decide at install OR update time, to use one or another of the sql?

But even if we do, remains the whole issue of a site moved from a 5.5.3+ server to a < 5.5.3 one and I believe that issue will be for all extensions out there.

richard67:
We should have done that differently in core for the extensions.

For a new install of the core, the **down**grading of sql statements from utf8mb4 to utf8 in joomla.sql works because new installer is running.

But for extensions we should have made a query **up**grade (which would have been a bit more sophisticated regarding replacement, but is possible), so extension still have their sqls in utf8 (but with new sizings of columns and indexes) and on a 3.5.0 or later, the database driver would upgrade these sql statements to utf8mb4 when installing/updating.

The only risk in using the larger columns sizes also on non-utf8mb4 is that if they use these larger sizes and then later migrate to utf8mb4, they might lose data because strings will be cut off at the end of varchar or char columns (tinytext, text and so on will be automatically enlarged to next size by mysql when conversting the table's or column's charset, so for these we keep the old sizes for utf8 and know they after conversion might be next larger size, e.g. tinytext will become text).

This is the only way I see making things work for extensions on pre 3.5.0 Joomla!s with non-utf8mb4 databases without having 2 variants for the extensions, like it was with 2.5 and 3.x.

Maybe we should discuss that with @wilsonge ? If we want to implement such a possibility to upgrade queries (or maybe even detect if to be upgraded or to be downgraded and then do whatever necessary), we might need that already for 3.5.

Or we rename 3.5 to 4.0 so extensions have to serve 2 versions, like it was with 2.5 and 3.x and the corresponding compability labels on the JED.

richard67:
@infograf768  Moving from a 5.5.3+ server where tables have utf8mb4 collations to a < 5.5.3 would not be possible without editing an sql dump to change charsets, and so this way is maybe simply not allowed. But moving from a < 5.5.3 to a 5.5.3+ should of course be possible, and then with the next update of the extension things should converted to utf8mb4.

richard67:
Maybe I should hide my profile picture and change my nick name so I will not be made responsible for this utf8mb4 conversion mess from now on until the end of my life (scared).

Or maybe it even needs a new identity for me to be really safe?

(sorry for silly jokes)

andrepereiradasilva:
@richard67 you have done a great job!

So to summarize the problematic question is: install/upgrade utf8mb4 extension in pre 3.5.0 Joomla in systems that use mysql database servers without utf8mb4 support (server version lower than 5.5.3).
- 3.5.0 seems all ok
- prior to 3.5.0 with mysql database servers with utf8mb4 support seems ok too (jean-marie will test better).

Bakual:
> Or we rename 3.5 to 4.0 so extensions have to serve 2 versions, like it was with 2.5 and 3.x and the corresponding compability labels on the JED.

Doesn't work, because in 4.0 it needs to be possible to have the same extension both in latest 3.x and 4.0. We can't break at will even with 4.0.

> remains the whole issue of a site moved from a 5.5.3+ server to a < 5.5.3 one 

That is something we don't support at all. Neither for PHP versions nor for MySQL. We only support upgrading, not downgrading.

As for extensions we need to make sure extensions written for 3.0 are still working in 3.5. We don't need to make sure that extensions written for 3.5 will still work in 3.0.
So an extension may opt to raise their minimum requirements and only support 3.5.0 and higher.

andrepereiradasilva:
one question since i don't quite yet understand how extensions install and upgrade work.

what does the `<file driver="mysql" charset="utf8">` is the extension manifest file does?
can it be set to utf8mb4 for instance?

See https://github.com/joomla-projects/com_localise/blob/develop/localise.xml#L17-L20

richard67:
Good question. Maybe just try it, have 2 files, one for utf8mb4 and utf8, and add those for utf8mb4 to the xml and see what happens?

infograf768:
I guess @Bakual answered our questions. :smile_cat: 
Let's keep com_localise as it is for now, and, if/when 4.0 is released, taken my age into account, we/someone can update the install.sql, if judged necessary.
After all 4 bytes characters are really rare as they are beyond the Plane.
see https://en.wikipedia.org/wiki/Plane_%28Unicode%29#Basic_Multilingual_Plane

andrepereiradasilva:
You can wait for 4.0, but don't think that's necessary, but it's your call.

I will leave this PR here and you guys merge it when you want.

It can also serve as a PoC for utf8 to utf8mb4 conversion in a extension with database tables for who needs.

richard67:
@andrepereiradasilva I you want I can help with a PoC. We should find out if and how the "charset=..." stuff in the XMLs is used by the core, and since when this is supported. Maybe it is a new thing which will not be present on older Joomla! versions.

andrepereiradasilva:
richard see https://github.com/joomla/joomla-cms/blob/staging/libraries/cms/installer/installer.php#L932

richard67:
@andrepereiradasilva that does not tell us since when it is there, and it just checks if uft8 or not, and if not has it empty. So it seems there is no way to use this to have separate versions for utf8mb4 and utf8, especially not on older installations where the installer does not have this code.

andrepereiradasilva:
> So to summarize the problematic question is: install/upgrade utf8mb4 extension in pre 3.5.0 Joomla in systems that use mysql database servers without utf8mb4 support (server version lower than 5.5.3).

Just tested it to see the actual result: Joomla 3.3.6 | MySql 5.1.73

#### New install
1. Install from URL: https://github.com/joomla-projects/com_localise/files/176570/com_localise-develop-utf8mb4.zip

Result:
image

#### Update

Update has a similar same result.
1. Install from URL: https://github.com/joomla-projects/com_localise/files/176570/com_localise-develop.zip
2. Upload: https://github.com/joomla-projects/com_localise/files/176570/com_localise-develop-utf8mb4.zip

Result:
image

richard67:
= Expected result :-(

andrepereiradasilva:
@richard67 do you have glip or something?

richard67:
What is glip?

andrepereiradasilva:
https://glip.com/

richard67:
No, I have skype only ... or Firefox' chat functions, but this I never used. Why? Something urgent? Will eat soon and then later maybe coud speak.

andrepereiradasilva:
no no not urgent, just a better way to talk

andrepereiradasilva:
removing my comments regarding this to not hijack the PR

richard67:
@andrepereiradasilva Anything new? Did you discuss that problem and this charset thing in the xml and how it is handled by the installer with someone? If you have skype and need to discuss you can get me there as richard.fath.67

infograf768:
BTw, I copied the wrong sql above.
It should be
`KEY `idx_path` (`path`(100))`
and not
`UNIQUE KEY `idx_path` (`path`(100))`

infograf768:
Ooops, I had pasted the correct one. (Me blind)

infograf768:
FYI: I fully tested last instance of com_localise (develop branch) with utf8mb4 sql on 3.4.8 and it looks that it is fully working. Mysql here is 5.5.25. There is no reason for this component to have mismatches.

andrepereiradasilva:
removed `UNIQUE` form `idx_path` index to avoid duplicate index problems, and resized the index to 191 for more consistency.

The 191 character limit is due to the maximum key length of 767 bytes. For a 4 byte character (utf8mb4), this means a max of 191 characters (floor(767/4) = 191).

Teste files updated

richard67:
@andrepereiradasilva Sounds good.

Other question: Do you think it could make sense to make the installer handle this charset info from the xml for charsets "utf8mb4" and "utf8" in following way:

If only 1 of these files is specified, on a Joomla! 3.5.0+ this file will be used and to be safe query downgrade will take place, which does not do any harm if utf8mb4 support and changes collations and charset from utf8mb4 to utf8 in queries if no utf8mb4 support. This will not work on pre 3.5 if the sql is made for utf8mb4 charset.

If both files are specified, on a  3.5.0+ the one is taken according to utf8mb4 support. No query downgrade needed but maybe could be done anyway without doing harm.

On a pre 3.5.0, the file for utf8mb4 will be not used because charset has to be hard-coded utf8.

So if you wanna have an extension which shall installable on pre 3.5.0 without utf8mb4 support and not be converted to utf8mb4 for 3.5.0+, use the utf8 file only, and this file may not use any utf8mb4 charsets or collations.

If it shall be installable on pre 3.5.0 without utf8mb4 support but also support utf8mb4 on 3.5.0+, use both files, and have the utf8 file using utf8 charsets and collations, and the utf8mb4 file using utf8mb4.

If it shall only be installable on pre 3.5.0, specify only 1 of the files (the utf8mb4 one if correct, but for backwards compatibility also utf8 would be used if utf8mb4 not specified). This 1 file is then made for utf8mb4.

What do you think about that? Could this be useful?

I would make a PR for core then, extending handling of charset info for install and uninstall sql files specified in xml manifests and should also add handling of charset field for update sql scripts.

andrepereiradasilva:
Your proposal will work in install/update extensions in all 3.x version and at the same time support both utf8 and utf8mb4 charsets in all mysql database environments?

If that is the case, IMHO it could be a good enhancement for supporting extensions in both utf8 (all versions) and utf8mb4 (3.5.0+) in extensions install/update.

At the same time simplifies the extensions creators job. Now for full support utf8mb4 in their extensions they will have to do a 3.5.0+ package and a pre 3.5.0 package or some custom install.

For what i understand in your proposal scenario one extension package can be made to work on all versions and support utf8mb4.

For what i understand your proposal is the 3.5.0 install process something like this:

``` xml
<install>
    <sql>
        <!-- utf8mb4 for Joomla 3.x versions with MySql 5.5.3+ -->
        <file driver="mysql" charset="utf8mb4">sql/install/mysql/install.sql</file>
        <!-- utf8 fallback for pre-3.5.0 versions with MySql lower than 5.5.3 -->
        <file driver="mysql">sql/install/mysql/install-utf8.sql</file>
    </sql>
</install>
<uninstall>
    <sql>
        <file driver="mysql">sql/install/mysql/uninstall.sql</file>
    </sql>
</uninstall>
<update>
    <schemas>
        <!-- utf8mb4 for Joomla 3.x versions with MySql 5.5.3+ -->
        <schemapath type="mysql" charset="utf8mb4">sql/updates/mysql</schemapath>
        <!-- utf8 fallback for pre-3.5.0 versions with MySql lower than 5.5.3 -->
        <schemapath type="mysql">sql/updates/mysql-utf8</schemapath>
    </schemas>
</update>
```

And since the pre 3.5.0 defaults are hardcoded with `utf8` they will only process the `utf8` (or empty) ones. The 3.5.0+ with your proposal will process the `utf8mb4` ones.

So like a utf8 fallback for pre 3.5.0 system.

But i don't have full knowledge how the extensions installer/updater works.

@wilsonge what do you think?

richard67:
@andrepereiradasilva Yes, this is what I meant.

Regarding having entries for mysql and mysqli: I have not looked into old versions down to 2.5, but currently it is sufficient to specify one of these 2 only, because for mysql, mysqli and pdomysql it all ends in mysql inside the php code before it is checked then. Maybe it is even bad to have entries for both, because it will then match 2 times, 1 time for mysql and 1 time for mysqli, and so run 2 times? Am not sure about the, maybe you or George can check or I can check later.

richard67:
Ah and you already saw that for uninstall no different versions are necessary. Another reason why I should make the core to be so tolerant and use the utf8 version then also for utf8mb4 (because only 1 specified). Or we have also versions for utf8mb4 there but pointing to the same file at the end, if we do not wanna make core be this tolerant but require the charset to fit exactly.

andrepereiradasilva:
so, for instance, if your solution was already working com_localise could work in all 3.x version with all supported mysql db versions with the xml posted in above comment and the necessary sql files.

IMHO it would be a good improvement.

updated the xml

richard67:
@andrepereiradasilva PR [#9468](https://github.com/joomla/joomla-cms/pull/9468) for the core is ready for code review, testing insructions will follow. If you want you can have a look on it and maybe even test with your test package for com_localise.

andrepereiradasilva:
updated test instructions

andrepereiradasilva:
@infograf768 with @richard67 joomla-cms new PR (https://github.com/joomla/joomla-cms/pull/9468) and the changes i made here this changes will work on 3.3+ versions and also on non utf8mb4 capable servers, so they there is no need to wait for 4.0 after.

richard67:
I tested this item :white_check_mark: successfully on [fd79cb3](https://github.com/joomla-projects/com_localise/pull/286/commits/fd79cb3d5575be7b14e132026f60fbd2717b1ef6) 

Tested all cases installation and then update to new version, uninstall this then, then install directly the new version, then uninstall this, with checking database after each step, for following Joomla! Version and utf8mb4 support combinations:
- 3.4.8 without utf8mb4 support
- 3.4.8 with utf8mb4 support
- 3.5.0 RC 4 + my PR [joomla/joomla-cms#9468](https://github.com/joomla/joomla-cms/pull/9468) without utf8mb4 support
- 3.5.0 RC 4 + my PR with utf8mb4 support

I checked data structures only, not if the localise component works well, because I am not familiar with the component.

Result: All as expected, i.e. on 3.5.0 RC 4 + my PR with utf8mb4 support tables are in utf8mb4, other cases tables are in utf8.

You should update your test instructions to make clear that on a 3.5.0 (+ my PR), the result is only utf8mb4_unicode_ci (and the 1 utf8mb4_bin column) when utf8mb4 is supported by the database server and client api, otherwise the result is like for the other case (3.4.8).

@infograf768 How can I mark a test result here without having a fancy issue tracker? I never marked test results with GitHub UI only so I have no idea. I helped myself by creating the above test result check mark and link to the commit manually.

andrepereiradasilva:
> You should update your test instructions to make clear that on a 3.5.0 (+ my PR), the result is only utf8mb4_unicode_ci (and the 1 utf8mb4_bin column) when utf8mb4 is supported by the database server and client api, otherwise the result is like for the other case (3.4.8).

updated

> @infograf768 How can I mark a test result here without having a fancy issue tracker? I never marked test results with GitHub UI only so I have no idea. I helped myself by creating the above test result check mark and link to the commit manually.

i think there is no issue tracker yet for com_localise
See https://issues.joomla.org/projects

infograf768:
For now, I have merged https://github.com/joomla-projects/com_localise/pull/289

When https://github.com/joomla/joomla-cms/pull/9468 will be merged, we can test this again after updating it.

infograf768:
btw, I do not think we need 191. It is used in core for some varchar. Nowhere in core do we use 191 for a KEY. It is always 100,

anibalsanchez:
Test OK

richard67:
@infograf768 You do not need 191, yes, but 191 is the correct value for max. possible.

I would also have used that in core instead of the 100 if the decision was on me.

But someone had a favour for even decimal numbers and so did not like the 191 before I started to work for the utf8mb4 stuff.

Is a bit like someone not familiar with dual system says: No I want this buffer be 1000 long and not 1024 because 1024 does not look nice :smile: 

But for real programmers the 1024 looks perfect, and so does the 191 for people knowing about mysql and utf8mb4.

andrepereiradasilva:
First of all let me say that i'm by no mean an expert in database systems, just trying to help the project.

Why the 191 index size?

In this case (the language file path index), when i added a size of 100 i noticed some values in the index were being cropped.
As i don't know exactly the consequences of the index values being cropped in this component (search, order, etc) i resized it to the maximum allowed value in utf8mb4 to prevent issues.

I was told that the maximum index length in utf8mb4 is 767 bytes (191 \* 4 bytes per character = 764), so there is the reason for the 191.

If anyone can confirm that having an index size of 100 will not cause issues, i will change it with no problems.

BTW fixed the conflicts after latest merge. Now the check for code difference is more clear.

infograf768:
As https://github.com/joomla/joomla-cms/pull/9468 is not going to make it in 3.5.0, I am going to release a 4.0.15-dev version of com_localise without this PR.

I am just waiting for this to be tested:
https://github.com/joomla-projects/com_localise/pull/290

richard67:
@infograf768 I agree, better wait with moving to utf8mb4 ... I think we can make the core handle conversion scripts for extensions, so extensions not have to do it with their schema updates or own php scripts (but still have to provide sql script(s) for conversion). Am almost sure I can have something ready soon, maybe for 3.5.1. What Ande does with this PR is a good proof of concept and still needed, only the conversion to utf8mb4 would then in a future solution be moved from the schemapath to a new conversionpath or so, and be executed by the installer on update or after migration when is necessary.

andrepereiradasilva:
closed as does not makes sense anymore

PR #287, opened 2016-03-17
Throw a warning when max_input_vars value is too low.

infograf768:
This PR, based partly on code from @Valc in https://github.com/joomla-projects/com_localise/pull/276
implements a Warning when the php.ini directive `max_input_vars` is lower than required to edit an ini file. It also prevents editing the file except under Source.

One should get:

screen shot 2016-03-17 at 16 37 42

The default on most systems is 1000 if the directive is not set at all.
To test, add 50 lines to admin en-GB.ini.
Modify the directive and test again.

The reason we can't use `$item->total` is that lines with comments are also counted as variables.


Valc:
Well done.
Now is easier to adjust if they leave more variables to consider, as we start taking the total number of lines in the file.

Regards.

infograf768:
Thanks for testing

PR #288, opened 2016-03-18
Add update tables

Valc:
Here there are the pending files for the simple develop feature.
Regards.


andrepereiradasilva:
@Valc that i know of you don't need the create table in every version, just the older one when the changes are introduced. I believe it is `4.0.0.sql`

Valc:
If you are sure, for me its ok.
The feature 'simple develop' will be released the next time, and will be introduced after the 4.0.14-dev version.
If such as you have leaved this one now, for example, an user is updating this component from him 4.0.11 version and the new table is added: perfect.
Regards.

andrepereiradasilva:
let me check that to be sure

So we should make it to 4.0.15 (since 4.0.14 is already released)

Bakual:
You don't need an SQL file for each version. Technically it doesn't even have to match the extension version.  Joomla just checks the #__schemas tables and applies each SQL file where the name is higher than the one stored in there.

andrepereiradasilva:
ok so this can be closed i already make a PR to set the update in 4.0.15 (https://github.com/joomla-projects/com_localise/pull/289)

infograf768:
Closing this in favour of #289

PR #292, opened 2016-03-21
Improve travis tests: test other php versions (5.6, 7 and hhvm); run code style tests only on one php version

andrepereiradasilva:
Check code difference

I'm not sure is anything else is needed.

the patchtester travis.yml is a little different reggarding CS
See https://github.com/joomla-extensions/patchtester/blob/master/.travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ matrix:
     - php: 5.5
     - php: 5.6
     - php: 7.0
+    - php: hhvm
+  allow_failures:
+    - php: hhvm
 
 before_script:
   # Make sure all dev dependencies are installed
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,6 @@ sudo: false
 
 language: php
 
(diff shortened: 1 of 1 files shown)


andrepereiradasilva:
it seems something more is needed because travis did not run on this PR

mbabker:
Since you're not actually using the `RUN_PHPCS` global you don't need to define it.  Otherwise, the test scripts are set up here differently than they are in the patchtester repo so not much else I can say on that.

andrepereiradasilva:
ok thanks @mbabker 
It's now testing on php 5.6 and php 7 too

andrepereiradasilva:
Guess what: it pass the php 7 test :)

infograf768:
I honestly have no idea if this should be merged or not.
@javigomez , please decide.

javigomez:
I see we are currently not running the system tests in travis so should not affect. Checking the builds after George's merge seems all to work okay.

Issue #297, opened 2016-03-25
EOL Issues with raw mode


Valc:
#### Steps to reproduce the issue

1 Edit any file with raw mode and save and close it.
2 Edit the same file in normal mode and save and close it.
3 Edit the same file in raw mode and go to the end of the file lines: before the last string now is added a section "[New strings]" when not true.
#### Expected result

It will be the same than edit in normal mode: the file is mounted with  "\n"
#### Actual result

Editing files in raw mode make save the files without Unix EOL, and one time the "BOM!" issues are inserted editing in raw mode then are making fail the normal mode saving files.
#### System information (as much as possible)

Linux, so my files are saved as default as UNIX/Linux EOL.
#### Additional comments

To catch the issue the MSP (Enabling the system messages options related with BOM and EOL) has been helping me to found whats was happening, due at the begin this was hard to catch.

The stuff is that edit in raw mode save the files without UNIX/Linux EOL, and my files are for sure UNIX/Linux EOL before edit it in raw mode.

After test the issue:
![eol](https://cloud.githubusercontent.com/assets/417247/14041098/1af30338-f26e-11e5-8525-28593062360c.png)

Regards.

infograf768:
I can't reproduce these issues here on Macintosh.

infograf768:
What I mean is that I can see your es-ES.com_ajax.ini is Windows EOL and es-ES.com_contacts.ini has [New Strings] on top of the file, but I can't reproduce your issue with my fr-FR files or Italian files

Valc:
Can you test saving one of you language files as UNIX/Linux EOL?

infograf768:
@Valc
Using 4.0.16-dev, I edit your 3.5.0.1 files (normal way) and save, and they are corrected:
no more Windows EOL for Ajax, no more [New Strings] on top for the other files concerned.

My files are always Unix.

infograf768:
Are'nt you using a custom com_localise and not 4.0.16-dev?

Valc:
No, i am updated today to .16. My other "complex develop" branch i know is bugged but i am waitting for more free time.
Noticed that saving in normal mode the issue about EOL is gone... But editing again in raw mode and saving for you is not coming back?

infograf768:
Nope.

infograf768:
wait, indeed it does saves with Windows EOL when saving in Raw mode.

Valc:
[...]

PR #298, opened 2016-03-26
Make sure EOL is Unix and not Windows (solves #297)

infograf768:
When saving a file using Raw mode (Source edit), the file end of lines are CRLF i.e. Windows, instead of LF i.e. Unix.

This comes from a flaw in CodeMirror.
This solves the issue for com_localise.


Valc:
It is working fine.

infograf768:
Thanks for testing. :smiley:

PR #299, opened 2016-03-27
override.ini should not be translated

infograf768:
In this case, override is the name of the file in the translations view.
It should never be translated.
Before patch, editing fr-FR

screen shot 2016-03-27 at 10 29 01

After patch

screen shot 2016-03-27 at 10 23 43


infograf768:
@Valc 
Thanks for testing.

Valc:
Working fine now :+1:

infograf768:
Thanks for testing.

PR #300, opened 2016-03-27
Add missed sections syntax

Valc:
I forguet add the [ ] when the sections line is mounted. This will solve my mistake.


infograf768:
@Valc
How do I test this?

infograf768:
OK, found. testing now.

Valc:
I start translating com_localise, that have a lot of sections, and then notice that when saved in normal mode, the sections syntax was crashed.

infograf768:
Confirmed and tested Ok.

Issue #302, opened 2016-03-28
Bacth process at packages view is crashing.


Valc:
#### Steps to reproduce the issue

Go to packages view and select any checkbox and then press the 'Batch' button.
#### Expected result

Access to "Batch" options.
#### Actual result

White page.
#### System information (as much as possible)

```
PHP instalado sobre:    Linux valc-MacBookPro 3.13.0-83-generic #127-Ubuntu SMP Fri Mar 11 00:26:47 UTC 2016 i686
Versión de la base de datos    5.5.47-0ubuntu0.14.04.1
Cotejamiento de la base de datos    utf8_general_ci
Cotejamiento de la conexión de la base de datos    utf8mb4_general_ci
Versión de PHP     5.5.9-1ubuntu4.14
Servidor web    Apache/2.4.7 (Ubuntu)
Interfaz de PHP     apache2handler
Versión de Joomla!     Joomla! 3.5.0 Stable [ Unicorn ] 21-March-2016 22:00 GMT
Versión de la plataforma de Joomla!    Joomla Platform 13.1.0 Stable [ Curiosity ] 24-Apr-2013 00:00 GMT
Agente del usuario  Mozilla/5.0 (X11; Ubuntu; Linux i686; rv:45.0) Gecko/20100101 Firefox/45.0 
```
#### Additional comments

Here the error log output:

```
[Mon Mar 28 17:17:34.777454 2016] [:error] [pid 1235] [client 127.0.0.1:37633] PHP Warning:  Missing argument 1 for JControllerForm::batch(), called in /var/www/html/joomla3/libraries/legacy/controller/legacy.php on line 728 and defined in /var/www/html/joomla3/libraries/legacy/controller/form.php on line 250, referer: http://localhost/joomla3/administrator/index.php?option=com_localise&view=packages
[Mon Mar 28 17:17:34.777807 2016] [:error] [pid 1235] [client 127.0.0.1:37633] PHP Notice:  Undefined variable: model in /var/www/html/joomla3/libraries/legacy/controller/form.php on line 274, referer: http://localhost/joomla3/administrator/index.php?option=com_localise&view=packages
[Mon Mar 28 17:17:34.777848 2016] [:error] [pid 1235] [client 127.0.0.1:37633] PHP Fatal error:  Call to a member function batch() on a non-object in /var/www/html/joomla3/libraries/legacy/controller/form.php on line 274, referer: http://localhost/joomla3/administrator/index.php?option=com_localise&view=packages
```

[...]

PR #303, opened 2016-03-29
Batch method is useless

infograf768:
Batch method is never used. See https://github.com/joomla-projects/com_localise/issues/302


Valc:
Ok, its not present yet.
Regards

Issue #305, opened 2016-04-22
Why the "filter" input box is CASE sensitive?


joomlagate:
#### Steps to reproduce the issue

Today I tried to translate an extension named Snippets with com_localise.
Snippets is downloaded from:
https://www.regularlabs.com/extensions/snippets

After installation of this extension, I went to com_localise backend, and choose client and language, then input "Snippets" into the filter input box of com_localise. Nothing found.

Then I input all-lower-case word "snippets", I got some results.
#### Expected result

I think this filter function should be CaSe InSenSitIVe. 
#### Actual result

But it is case insensitive, only allows complete lower case strings.
#### System information (as much as possible)

Joomla 3.5.1
PHP 7
com_localise 4.0.16 dev
#### Additional comments

Could you please let this filter function to be CaSe InSenSitIVe so that I can just copy the extension's name from that webpage and then paste it into the input box to search it, in this way, I think it can boost my effeciency than input a complete lower case sting MANUALLY.

Thank you.

infograf768:
@baijianpeng 
Can you test by replacing the file
`ROOT/administrator/components/com_localise/models/translations.php`
by this one (after decompressing):

[translations.php.zip](https://github.com/joomla-projects/com_localise/files/232993/translations.php.zip)

joomlagate:
Yes, this modified version works fine, now I can filter with any CasE
combination.

Thank you.

白建鹏

https://www.joomlagate.com/

微信公众号：joomlagate
新浪微博： @joomlagate

---

2016-04-23 22:47 GMT+08:00 infograf768 notifications@github.com:

infograf768:
Great, please add Test OK here:
https://github.com/joomla-projects/com_localise/pull/306

infograf768:
Closing as we have a patch.

PR #306, opened 2016-04-23
Allowing case insensitive search in translations view

infograf768:
Solving https://github.com/joomla-projects/com_localise/issues/305
The patch gets rid of eventual spaces and changes search state to lowercase.
Also added a message when filtering gets no result.


joomlagate:
My test proves that this new PR solved #305 problem.

Thank you.

(I did not find a button "test OK" so that I just write some text here, is it right?)

infograf768:
Perfect.

PR #307, opened 2016-04-26
[NEW] Allowing copy reference language to new language

infograf768:
Until now, when one wants to create a new core language in com_localise, one has first to create the xmls for admin and site, then go to the translations page and edit/save each file. Then follow the usual way to create an installable pack.
This is a long process, which is worth it when a real translation is needed.
But when one does only need to use an already existing translation with a different country code, it takes ages...

This PR is proposed to make this task much faster.
One will for example be able to fastly create a fr-CH pack when the French language is targetting Switzerland, or en-JM pack when targetting Jamaica, etc.
This is specially useful for a multilanguage site.

To test, install any existing language pack or just use en-GB if what you need is an English base language.

Set in the component Options the Reference language you need to copy from.
Then go to the languages page and create a new language. Remark the tip on top of the page

localise_jamaica_4

Let's create a new language en-JM for Jamaica and save (here for the site client). The reference language is set to en-GB. remark the new Toolbar button and the tip on top of the page.

screen shot 2016-04-26 at 16 54 17

If we just Close, the new language appears in the Languages page:

localise_jamaica_1

Before this PR, if we go to the Translations page, we would get 

localise_jamaica_translations_before

Remark  the Translated column where all bars are red and 0%. This means the files for en-JM are NOT yet created. They would need to be edited and saved to exist in the en-JM folder.

Let's go back to Languages and edit the English (Jamaica) file. 
Then click on the button "Copy reference language files to this language".

If no errors, for example, if there is ONLY the en-JM.xml file in the en-JM folder, the process will copy and save the new en-JM.XXX.ini files in the en-JM folder for this client. It will also copy and modify the en-JM.localise.php.

localise_jamaica_3

Now if we go to the Translations page, we will get:

localise_jamaica_translations

Which shows the files are present and all strings are the same as the original reference language ones.

Then it will only be a matter of following the Tip advice to create a new pack for en-JM

```
COM_LOCALISE_COPY_REF_TO_NEW_LANG_TIP="Clicking the <strong>Copy reference language files to this language</strong> toolbar button will copy all reference language files to this new language for this client. The new language folder should only contain this <strong>%1$s.xml</strong> file.<br />The Reference Language should be defined in the Component Options. It is presently set to <strong>%2$s</strong>.<br />Do this for both administrator and site. Then go to the Translations page to modify translations if necessary.<br />Thereafter, you may create a fully working installation pack for this language by navigating to the Packages page, create a new package by duplicating a master core pack, modify to fit the new language, download and install via Extensions=>Install.<br />This makes it easier to create custom language packs for various countries than saving each file in the Translations page.<br /><strong>Examples of use:</strong> create an <strong>en-JM</strong> pack targeting Jamaica from the en-GB reference language, a <strong>fr-CH</strong> pack targeting Switzerland from the fr-FR reference language, etc." 
```

Any advice on possible improvements to this PR are welcome.

@Valc @mbabker @Bakual @roland-d


infograf768:
Hmm, found a bug. This will work correctly ONLY if the Filter in the Languages page is set to the specific client and new language. I guess because of
`$data    = $app->getUserState('com_localise.select');`
as tag as well as client are fetched from $data.

infograf768:
So, needs some work. :)

infograf768:
Should be OK now. Please test. @Valc

andrepereiradasilva:
Test OK.

Valc:
Good idea!
It's working fine here.
Regards.

infograf768:
Thanks for testing. Merged and preparing new release.

Issue #309, opened 2016-04-29
how to hide the "check if you set complete" pop-up warning?


joomlagate:
#### Steps to reproduce the issue

Each time I save the translation in com_localise, after click on the Save button, there will show a pop-up dialog remind me to check if I had set "complete" on the details Tab.

This pop-up is annoying, it forces me to click the "OK" button on that pop-up message,  but there is no option to disable or hide it.
#### Expected result

I hope there is an option to allow me totally hide that pop-up message.
#### Actual result

I did not find any option to do so.
#### System information (as much as possible)

Joomla  3.5.1
Localise 4.0.17 dev
#### Additional comments

Thank you.

infograf768:
Hmm, this imply getting the php parameter into js.

infograf768:
I can do that, but, again, another parameter...

infograf768:
@Valc 
The code is ready here (quite simple). Do you agree I add this parameter in the component Options?

Valc:
I think that if you have to include it as an option it should be enabled by default, due for me that reminder is very useful (It saves me time and mistakes). But, if other persons does not like use the popup window they can go to disable it by Options and also can get their taste with that popup.

Regards, Valc.

infograf768:
Sure, would be enabled by default.
Will make a PR

infograf768:
Please test https://github.com/joomla-projects/com_localise/pull/310

PR #310, opened 2016-05-01
New parameter to display an alert or not when saving translation

infograf768:
Pull Request for Issue https://github.com/joomla-projects/com_localise/issues/309

After patch, a new parameter is available in the component Options. Default is Yes.
screen shot 2016-05-01 at 18 11 22


joomlagate:
It is difficult to download the “changed files" from Github. Why not put a download button there?

I prefer to wait for next release to see the change.

Valc:
Test ok.

infograf768:
Thanks for testing. :)

PR #322, opened 2017-12-19
Composer

Bakual:
@infograf768 Can you test if that works?
It would add the Joomla/Github package from Composer and load it through the composer autoloader.
I haven't tested it yet myself, just threw this together 😄 

I did it using a second composer.json file because we need the classes in the component directory.


infograf768:
Needs some small changes to use in 3.x.
require_once JPATH_ADMINISTRATOR . '/components/com_localise/vendor/autoload.php';
to be added in models/fields/releases.php

For 4.0 we will need to completely refactor the component as I have many errors.

Bakual:
Added the autoloader to JFormFieldReleases.
As for 4.0, many changes can already be done in J3.x. It will raise the minimum required Joomla version to J3.8 thought.

Bakual:
There is something I dislike about this way. Due to how composer works, it will load all dependencies of Joomla/Github as well. But those are already present in the Joomla CMS vendor folder. So we're duplicating those classes here;:
````
    "ircmaxell/password-compat": "v1.0.4",
    "joomla/compat": "1.2.0",
    "joomla/registry": "1.5.2",
    "joomla/string": "1.4.1",
    "joomla/uri": "1.1.1",
    "joomla/utilities": "1.4.1",
    "symfony/polyfill-php55": "v1.6.0"
````

From reading, it should be possible to declare those in a "provide" block in the composer.json file, however I couldn't make it work so composer wouldn't pull those classes in.

infograf768:
`As for 4.0, many changes can already be done in J3.x. It will raise the minimum required Joomla version to J3.8 thought.`

Indeed, we could add an alert. Looking into it now.

> There is something I dislike about this way. Due to how composer works, it will load all dependencies of Joomla/Github as well.

That is over my skills but I see that the new vendor folder is a huge 1,8 Mb on disk... for 211 items...

infograf768:
I guess missing in patch the vendor folder in localise.xml.
`<folder>vendor</folder>`

Bakual:
True, I have to add that folder into the XML.
And yes, the dependencies of the GitHub are crazy. Technically it would be no issue since most of them are already in Joomla available, but if we have to ship it as part of com_localise due to the way composer works, it pretty much makes composer unuseable for our needs. I'll ask around if someone knows a solution to this.

Bakual:
Added the vendor folder to the XML.

Bakual:
@infograf768 Figured it out with some help. Now it only contains the stuff which isn't already in Joomla.

infograf768:
Test looks OK here on 3.8.4
vendor folder went down
`Size: 918 053 bytes (1,1 MB on disk) for 113 items`

infograf768:
Note
> Figured it out with some help

may I suggest to post somewhere a tutorial as this would be extremely useful for 3rd party devs?

infograf768:
Thanks. Will now take care of min version.

Bakual:
> may I suggest to post somewhere a tutorial as this would be extremely useful for 3rd party devs?

I wouldn't know where to do that so it would be found by developers. Basically it would be this lines in the composer.json:
https://github.com/joomla-projects/com_localise/blob/7799bc4c5e84c9d35a8a3414797146d8afa8b765/component/composer.json#L14-L22

Now composer will not install the mentioned packages as dependencies. It's a hack but works.

PR #323, opened 2017-12-21
Raise minimum version to 3.8.0

Bakual:
Since we start using namespaced classes with the GitHub package we need to ensure com_localise is only installed in J3.8.0 and higher.

This PR raises the minimum version and changes the scriptfile to use the existing checks in "InstallerScript"

Now I'm not sure if we should use the old `JInstallerScript` or already the new namespaced variant like in thsi PR. In J3.8.0 both will work. In J3.7.x only the old will work and in 4.0 only the new one.
So it's a question where the installation should break. 😄

### DIFF
--- a/install.php
+++ b/install.php
@@ -9,41 +9,35 @@
 
 defined('_JEXEC') or die;
 
+use Joomla\CMS\Installer\InstallerScript;
+
 /**
  * Installation class to perform additional changes during install/uninstall/update
  *
  * @package  Localise
  * @since    4.0
  */
-class Com_LocaliseInstallerScript
+class Com_LocaliseInstallerScript extends InstallerScript
 {
 	/**
-	 * Minimum supported version of the CMS
+	 * The extension name. This should be set in the installer script.
 	 *
 	 * @var    string
-	 * @since  4.0
+	 * @since  4.0.31
 	 */
-	protected $minCmsVersion = '3.3';
-
+	protected $extension = 'com_localise';
 	/**
-	 * Function to act prior to installation process begins
+	 * Minimum PHP version required to install the extension
 	 *
-	 * @param   string               $type    The action being performed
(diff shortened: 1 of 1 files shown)


infograf768:
@Bakual 
IMHO we should display an alert if Joomla is < 3.8 as the new Github code is supposed to work only in >= 3.8.0
Therefore the code used should work both in 3.8, 3.9 and 4.0 and the alert should display when installing, as this PR does.
The string used though is not very nice when using `protected $minimumJoomla = '3.8.0';` which will give
`JLIB_INSTALLER_MINIMUM_JOOMLA="You don't have the minimum Joomla version requirement of J%s"` (Imho not very nice English)
instead of what we have now.
`COM_LOCALISE_ERROR_INSTALL_JVERSION="The Localise component requires version %s or newer of Joomla!"`

In any case, we are going to have issues with com_localise and 4.0 as a lot of the code will have to be modified, including display. We will have to devise a way to propose 2 versions of the component. We can talk about that later.

Merging and thanks. my PR ready to prevent an already installed component to display its Managers if J version is < 3.8

infograf768:
See https://github.com/joomla-projects/com_localise/pull/324

wilsonge:
FWIW given the use case here I'd use the non-namespaced version of InstallerScript give yourself some defined errors on old J Versions

Bakual:
@wilsonge Do you think we can keep the alias for InstallerScript in 4.0? Assuming my extension works from 3.8.0 upwards and also in 4.0.0 (using some switches for eg layouts), then I can't use InstallerScript for this as it breaks in <3.8.0. GIven it is supposed to give an error message in <3.8.0, this is a bit less than optimal 😄

Bakual:
Sorry, forget what I just wrote. I just saw it is only deprecated for 5.0. I just assumed the class maps are going to be removed with 4.0.
Will make a PR to change it to JInstallerScript.

PR #325, opened 2017-12-26
Use JInstallerScript instead of the namespaced counterpart

Bakual:
As suggested by @wilsonge in https://github.com/joomla-projects/com_localise/pull/323#issuecomment-353978047


infograf768:
OK

